PR #11, opened 2025-01-23
cmake: arm: linker Added LMA support for dts memory section

rajranjan-alifsemi:
In DTS memory section LMA as optional parameter has been included.


VeijoPesonen (changes requested):
You need to create a PR to alifsemi/sdk-alif which takes these changes into use.

PR #36, opened 2025-02-07
dts: arm: alif: ensemble: Remove reference to CSI bus node

ChandraBhushan-AlifSemi:
Moved the reference to the CSI bus node in the camera controller node from the SoC DTS to the ARX3A0 camera sensor overlay, as we need a reference to this bus only in the case of serial camera sensors.


ChandraBhushan-AlifSemi:
It is taken back in use in video sample in the following pull request :- https://github.com/alifsemi/sdk-alif/pull/31

HemanthKumar-AlifSemi (approved):

Prabhakar-Alifsemi (approved):

PR #39, opened 2025-02-10
drivers: sensor: icm42670: update sensor startup-time delay

Shreehari-AlifSemi:
1. Updated sensor startup time delay
2. Removed Dbg log config from project configuration


VeijoPesonen (approved):

YogenderKumar-AlifSemi (approved):
Looks Good!

PR #48, opened 2025-02-18
dts: arm: Remove pinctrl from GPIOs in Ensemble

Prabhakar-Alifsemi:
By default, all GPIOs are enabled, and during the pre-kernel stage, the GPIO driver sets all pinmux configurations to default (GPIO). This causes issues when multiple cores are running. For example, if the first core runs I2C and the second core runs SPI, the following sequence occurs:
- The first core sets the pinmux to I2C.
- The second core starts and sets all pins to their default (GPIO). This results in the pinmux configuration set by the first core being overwritten by the second core, breaking the first core's I2C functionality, while the second core works properly. To fix this issue, the default GPIO pinctrl property is removed.

### DIFF
--- a/dts/arm/alif/ensemble_rtss_common.dtsi
+++ b/dts/arm/alif/ensemble_rtss_common.dtsi
@@ -587,8 +587,6 @@
 			     <181 0>, <182 0>,
 			     <183 0>, <184 0>,
 			     <185 0>, <186 0>;
-		pinctrl-0 = < &pinctrl_gpio0 >;
-		pinctrl-names = "default";
 		gpio-controller;
 		#gpio-cells = <2>;
 	};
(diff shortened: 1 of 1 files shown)


sudhir-alifsemi (dismissed):

silesh-alifsemi (dismissed):
Look OK. Should we also mark all the gpio nodes as "disabled" here and then enable them on need basis (from the board dts for example) ?

silesh-alifsemi (changes requested):
"dts: arm: alif: ensemble: remove pinctrl info from gpio nodes" is a better subject line for this commit.

silesh-alifsemi (approved):

sudhir-alifsemi (approved):

PR #49, opened 2025-02-18
drv: ospi-flash: removed warnings

GaneshRamani-AlifSemi:
ARM-GCC compiler treating off_t as long int type and Arm-clang treat this is uint32_t.

As per Zephyr it should be treated as integer pointer.

Zephyr [off_t reference](https://docs.zephyrproject.org/latest/doxygen/html/lib_2libc_2minimal_2include_2sys_2types_8h.html#a98a5cc5f4a350bf5652565021a2f239f)

### DIFF
--- a/drivers/flash/flash_ospi_is25wx.c
+++ b/drivers/flash/flash_ospi_is25wx.c
@@ -224,7 +224,7 @@ static int flash_alif_ospi_read(const struct device *dev, off_t address, void *b
 	dev_data->trans_conf.wait_cycles = 0;
 	dev_data->trans_conf.addr_len = 0;
 
-	LOG_DBG("read address %ld length to read %d", address, length);
+	LOG_DBG("read address %u length to read %d", (uint32_t) address, length);
 
 	ret = alif_hal_ospi_prepare_transfer(dev_data->ospi_handle, &dev_data->trans_conf);
 	if (ret != 0) {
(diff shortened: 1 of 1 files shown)


drivers/flash/flash_ospi_is25wx.c
```diff
@@ -224,7 +224,7 @@ static int flash_alif_ospi_read(const struct device *dev, off_t address, void *b
 	dev_data->trans_conf.wait_cycles = 0;
 	dev_data->trans_conf.addr_len = 0;
 
-	LOG_DBG("read address %ld length to read %d", address, length);
+	LOG_DBG("read address %d length to read %d", (uint32_t) address, length);
```
AkiMakkonen-AlifSemi, line 227: Is there a reason why you are using %d (signed integer) for unsigned type? Type should be %u? Or even better you could print out a hexadecimal address for enhanced readability.

GaneshRamani-AlifSemi: Thanks for pointing out, it is offset don't want to confuse as absolute address, now I changed to %u.

AkiMakkonen-AlifSemi: Ok, I see. Thanks.

AkiMakkonen-AlifSemi (approved):

PR #67, opened 2025-03-25
Flash shell support for IS_Flash device.

GaneshRamani-AlifSemi:
Supporting Flash access through Shell module.

+ Added 8-bit buffer R/W changes as required for shell module to read and write the content to Flash.
+ Renamed to Flash controller node. OSPI driver not exposed from Zephyr and it implemented through HAL.
+ Flash properties updated and Storage node added.
+ Renamed macros to be generic.

### DIFF
--- a/drivers/flash/flash_ospi_is25wx.c
+++ b/drivers/flash/flash_ospi_is25wx.c
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#define DT_DRV_COMPAT snps_desigware_ospi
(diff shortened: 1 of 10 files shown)


jheiskan81 (approved):
Is e7 only user for shell sample?

jheiskan81 (approved):

AkiMakkonen-AlifSemi (approved):

jheiskan81 (changes requested):
take commits from #85 because it touch to same driver

GaneshRamani-AlifSemi:
this can be closed as PR merged with #88

PR #68, opened 2025-03-25
drivers: serial: ns16550: Don't flush the FIFO only if a device is used as a wakeup source.

VeijoPesonen:
By default we want to flush the FIFOs if the UART instance are not used as a wakeup sources. When acting as a wakeup source it's required to keep the FIFO content so that no data is lost.

### DIFF
--- a/drivers/serial/uart_ns16550.c
+++ b/drivers/serial/uart_ns16550.c
@@ -47,6 +47,7 @@ LOG_MODULE_REGISTER(uart_ns16550, CONFIG_UART_LOG_LEVEL);
 #define UART_NS16550_PCP_ENABLED DT_ANY_INST_HAS_PROP_STATUS_OKAY(pcp)
 #define UART_NS16550_DLF_ENABLED DT_ANY_INST_HAS_PROP_STATUS_OKAY(dlf)
 #define UART_NS16550_DMAS_ENABLED DT_ANY_INST_HAS_PROP_STATUS_OKAY(dmas)
+#define UART_NS16550_FIFO_PRESERVE_ENABLED DT_ANY_INST_HAS_PROP_STATUS_OKAY(fifo_preserve)
 
 #if DT_ANY_INST_ON_BUS_STATUS_OKAY(pcie)
 BUILD_ASSERT(IS_ENABLED(CONFIG_PCIE), "NS16550(s) in DT need CONFIG_PCIE");
@@ -360,6 +361,9 @@ struct uart_ns16550_dev_data {
 	DEVICE_MMIO_RAM;
 	struct uart_config uart_config;
 	struct k_spinlock lock;
+#if UART_NS16550_FIFO_PRESERVE_ENABLED
+	bool fifo_preserve;
+#endif
 	uint8_t fifo_size;
 	uint16_t irq;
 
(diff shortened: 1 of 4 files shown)


drivers/serial/uart_ns16550.c
```diff
@@ -698,9 +702,15 @@ static int uart_ns16550_configure(const struct device *dev,
 	static uint32_t resuming __attribute__((noinit));
 	static const int resume_magic = 0x45454545;
 
-	/* Keep FIFOs' content */
 	if (resuming == resume_magic) {
-		fifo_clr_mask = 0;
+#if UART_NS16550_FIFO_PRESERVE_ENABLED
```
KimmoVaisanen-AlifSemi, line 706: `UART_NS16550_FIFO_PRESERVE_ENABLED` is now a bit misleading. I think this is now something like "force FIFO clear in resume".

sudhir-alifsemi (commented):
In PM for device, zephyr has a property called "wakeup-source" which declares that this device can be used as a wakeup source. So, I was thinking whether we should have a flag just for preserving the FIFO or can we use this property to achieve the desired implementation.

mikter:
> In PM for device, zephyr has a property called "wakeup-source" which declares that this device can be used as a wakeup source. So, I was thinking whether we should have a flag just for preserving the FIFO or can we use this property to achieve the desired implementation.

I think this is a good idea. Also one thing for wakeup source is that when to enable IRQ and call the user callback it should not be done before it is set so perhaps that is also needed

VeijoPesonen:
> In PM for device, zephyr has a property called "wakeup-source" which declares that this device can be used as a wakeup source. So, I was thinking whether we should have a flag just for preserving the FIFO or can we use this property to achieve the desired implementation.

Actually the directive works as an advice for the PM framework itself so using it as an hint to preserve FIFO is wrong from my point of view. I mean, if I enable the flag it's not only preserving FIFO what needs to be done.

* [Wakeup capability](https://docs.zephyrproject.org/3.7.0/services/pm/device.html#wakeup-capability)
"""
This property is only used by the system power management to identify devices that should not be suspended. It is responsibility of driver or the application to do any additional configuration required by the device to support it.
"""

We wouldn't be using the property as intended as the expectation is to do warm boot and reinitializing the driver. It has nothing to do with suspension of peripherals. That's why I would prefer to keep the implementation as it is.

drivers/serial/uart_ns16550.c
```diff
@@ -2012,6 +2024,8 @@ static int uart_ns16550_pm_action(const struct device *dev, enum pm_device_actio
 			COND_CODE_1(DT_INST_PROP_OR(n, hw_flow_control, 0),          \
 				    (UART_CFG_FLOW_CTRL_RTS_CTS),                    \
 				    (UART_CFG_FLOW_CTRL_NONE)),                      \
+		IF_ENABLED(DT_INST_NODE_HAS_PROP(n, fifo_preserve),                  \
```
petrih-alifsemi, line 2027:
```suggestion
```
`DT_INST_PROP_OR` is handling also the missing property.

petrih-alifsemi: ...or do you mean `UART_NS16550_FIFO_PRESERVE_ENABLED`?

PR #69, opened 2025-03-27
dts: arm: alif: Ensemble/Balletto: Update NVIC IRQ priority bits to 8

Tanay-AlifSemi:
- Set NVIC num-irq-priority-bits to 8 for both Ensemble and Balletto families.
  - For Ensemble, relocated irq-priority-bits from board files to DTS.
- Updated default IRQ priority level to 0 (to match with CMSIS pack).
https://alifsemi.atlassian.net/browse/ZRTSS-480

### DIFF
--- a/dts/arm/alif/balletto_rtss_common.dtsi
+++ b/dts/arm/alif/balletto_rtss_common.dtsi
@@ -62,25 +62,25 @@
 			seservice0r: mhu@40040000 {
 				compatible = "arm,mhuv2";
 				reg = < 0x40040000 0x1000 >;
-				interrupts = < 37 3 >;
+				interrupts = < 37 0 >;
 				interrupt-names = "rx";
 			};
 			seservice0s: mhu@40050000 {
 				compatible = "arm,mhuv2";
 				reg = < 0x40050000 0x1000 >;
-				interrupts = < 38 3 >;
+				interrupts = < 38 0 >;
 				interrupt-names = "tx";
 			};
 			seservice1r: mhu@40060000 {
 				compatible = "arm,mhuv2";
 				reg = < 0x40060000 0x1000 >;
-				interrupts = < 39 3 >;
+				interrupts = < 39 0 >;
 				interrupt-names = "rx";
 			};
 			seservice1s: mhu@40070000 {
 				compatible = "arm,mhuv2";
 				reg = < 0x40070000 0x1000 >;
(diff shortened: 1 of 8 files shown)


KarolSaja-AlifSemi (approved):

kjbracey (dismissed):
I'm actually not a huge fan of the CMSIS pack defaulting to zero - a mid value makes more sense.

I'd be inclined to match typical Zephyr behaviour from other vendors rather than the CMSIS pack on this case - interrupt priority techniques are OS/environment specific, rather than hardware specific.

Tanay-AlifSemi:
> I'm actually not a huge fan of the CMSIS pack defaulting to zero - a mid value makes more sense.
> 
> I'd be inclined to match typical Zephyr behaviour from other vendors rather than the CMSIS pack on this case - interrupt priority techniques are OS/environment specific, rather than hardware specific.

NXP/ST keeps default irq priority level as 0.

NXP https://github.com/alifsemi/zephyr_alif/blob/main/dts/arm/nxp/nxp_imx6sx_m4.dtsi
image

ST https://github.com/alifsemi/zephyr_alif/blob/main/dts/arm/st/c0/stm32c0.dtsi
image

sudhir-alifsemi (approved):

jheiskan81 (dismissed):

VeijoPesonen:
@Tanay-AlifSemi Please rebase and fix the conflicts. Please create a PR to alifsemi/sdk-alif to update the west manifest.

dts/arm/alif/balletto_rtss_common.dtsi
```diff
@@ -54,7 +54,7 @@
 			ethosu0: ethosu55@400E1000 {
 				compatible = "arm,ethos-u";
```
jheiskan81, line 55: Where is coming that ethou0: ethosu55@400e100 definition.
originally there was not a defined region so i move that outside from soc:
https://github.com/alifsemi/zephyr_alif/commit/c63e700e168493b0fd2d4f89843ddcf3274d5320. 
If i looking for file history at github I dont see these defitions at all?

Tanay-AlifSemi: fixed conflict.

jheiskan81 (approved):

kjbracey (approved):

PR #74, opened 2025-04-02
soc: arm: alif_ensemble alif_balletto: OSPI XIP memory region handling

AkiMakkonen-AlifSemi:
Added OSPI[0|1] linker sections.
Added generation of separate .bin files for external flash and MRAM.

### DIFF
--- a/soc/arm/alif_ensemble/common/Kconfig.soc
+++ b/soc/arm/alif_ensemble/common/Kconfig.soc
@@ -89,3 +89,15 @@ config  SRAM1_REGION
 	default "SRAM1" if $(dt_nodelabel_enabled,sram1)
 	help
 	    Provide config SRAM1 region
+
+config OSPI0_XIP_REGION
+	string
+	default "ospi0" if $(dt_nodelabel_enabled,ospi0)
+	help
+	    Configure OSPI0 XIP memory mapped region name
+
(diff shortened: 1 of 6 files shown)


AnttiKauppila-AlifSemi (approved):

soc/arm/alif_ensemble/common/Kconfig.soc
```diff
@@ -89,3 +89,15 @@ config  SRAM1_REGION
 	default "SRAM1" if $(dt_nodelabel_enabled,sram1)
 	help
 	    Provide config SRAM1 region
+
+config OSPI0_XIP_REGION
+	string
+	default "ospi0" if $(dt_nodelabel_enabled,ospi0)
```
rajranjan-alifsemi, line 95: this default option is conditional, if that condition did not meet,have that path also  to through proper error in cmake

rajranjan-alifsemi (changes requested):

soc/arm/alif_ensemble/common/Kconfig.soc
AkiMakkonen-AlifSemi: I'm sorry I don't understand your point. Please clarify. It is not an error case if the ospi0 is not enabled so it should not give error.

PR #85, opened 2025-04-10
OSPI clk frequency support

GaneshRamani-AlifSemi:
Updated ospi clk frequency from hardcoded.
Clock Frequency shall be retrieved from OSPI node.

### DIFF
--- a/drivers/flash/flash_ospi_is25wx.c
+++ b/drivers/flash/flash_ospi_is25wx.c
@@ -648,7 +648,7 @@ static int flash_alif_ospi_init(const struct device *dev)
 
 	memset(&init_config, 0, sizeof(struct ospi_init));
 
-	init_config.core_clk = SYS_AXI_CLK;
+	init_config.core_clk = DT_PROP(ALIF_OSPI_NODE, clock_frequency);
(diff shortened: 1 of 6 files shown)


jheiskan81 (approved):
Looks good for me

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -648,7 +648,7 @@ static int flash_alif_ospi_init(const struct device *dev)
 
 	memset(&init_config, 0, sizeof(struct ospi_init));
 
-	init_config.core_clk = SYS_AXI_CLK;
+	init_config.core_clk = DT_PROP(ALIF_OSPI_NODE, clock_frequency);
```
jheiskan81, line 651: Minor question is this driver generic? if it is we can't use ALIF_OSPI_NODE

AkiMakkonen-AlifSemi (approved):

jheiskan81:
#67 what is difference between these? You are touching same files

jheiskan81:
this PR commit could be added to #67 ? i think it make sense

PR #90, opened 2025-04-11
build: automation: jenkins file update

rajranjan-alifsemi:
removed docker build and setup initial build steps.

### DIFF
--- a/automation/Jenkinsfile
+++ b/automation/Jenkinsfile
@@ -1,55 +1,131 @@
-def BRANCH = "main"
-
 pipeline {
     agent any
     options {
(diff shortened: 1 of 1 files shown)


VeijoPesonen (approved):

sudhir-alifsemi (approved):
I would prefer to have VM(docker env) to run this build against the zephyr required host tools(ubuntu version, python, cmake version etc) against the zephyr version. We can run into issues otherwise.

But this can be worked upon later.

AnttiKauppila-AlifSemi (approved):

PR #91, opened 2025-04-11
Fixes for mcuboot/mcumgr 

silesh-alifsemi:
Add` zephyr,uart-mcumgr` chosen nodes for the ensemble e3/e7 devkits. Increase the application image slot sizes for mcuboot compatible application builds. Also, move the mram storage partitions from the board device tree files into the HP/HE core device tree include files as that is where they really belong. While increasing the image slot sizes, care has been taken to not to overflow into the mcuboot/non-mcuboot application builds for the other core (The current `FLASH_LOAD_OFFSET` for HP is at 2MB).

### DIFF
--- a/boards/arm/alif_e3_dk/alif_e3_dk_rtss_he.dts
+++ b/boards/arm/alif_e3_dk/alif_e3_dk_rtss_he.dts
@@ -19,6 +19,7 @@
 		zephyr,sram = &dtcm;
 		zephyr,console = &uart4;
 		zephyr,shell-uart = &uart4;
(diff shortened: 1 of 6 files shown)


silesh-alifsemi:
> Do we need same for B1 and E1C?

I do not think the` mram_storage` changes are needed for B1 and E1C as the slots are already big enough to accommodate reasonably sized application images. But we will need the` zephyr,uart-mcumgr` chosen nodes. Will raise another PR for that.

RupeshKumar-AlifSemi (approved):
Looks OK to me

PR #99, opened 2025-04-16
drivers: I2C_DW: Fix I2C scan example

Shreehari-AlifSemi:
Solves two identical issues listed below:

Issue 1: I2C scanner example for DesignWare hardware gets stuck indefenitely resulting in system hang up.This is because DW I2C driver does not handle 0 byte transfer correctly which is the case for I2C scan example.

Fixed it by overwriting the msg length to 1 if it is 0 and the buffer is not NULL.

Issue 2: Similarly, if the I2C pins are not pulled up (nothing connected
 to I2C pins), the DW hardware does not actually send the data
(assuming contention on the bus) hence not releasing the semaphore resulting in calling thread waiting forever.

Fixed it by adding a timeout to k_sem_take call and return error if cannot successfully acquire it.

Tested scenarios where nothing was connected on the bus and saw the I2C scan example complete the whole scan command. Then connected two different sensors on the I2C bus and saw both the address on the console. Tested both the uses cases on Raspberry Pi Pico.

Fixes #70332.

Found that micropython tackles the same issue by implementing I2C scan commands with Soft I2C because the same reason mentioned in Issue 1.

### DIFF
--- a/drivers/i2c/Kconfig.dw
+++ b/drivers/i2c/Kconfig.dw
@@ -21,3 +21,8 @@ config I2C_DW_LPSS_DMA
 	  This option enables I2C DMA feature to be used for asynchronous
 	  data transfers. All Tx operations are done using dma channel 0 and
 	  all Rx operations are done using dma channel 1.
+
+config I2C_DW_RW_TIMEOUT_MS
+	int "Set the Read/Write timeout in milliseconds"
+	depends on I2C_DW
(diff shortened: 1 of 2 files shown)


sudhir-alifsemi (approved):

RupeshKumar-AlifSemi (approved):
Looks OK

PR #103, opened 2025-04-18
QDEC driver support

Manoj-AlifSemi:
UTIMER QDEC driver support.
Depends on alifsemi/hal_alif#29
child node counter is added to all existing UTIMER nodes across the soc families which is used in the utimer counter driver. Changes are made accordingly to dependent files.


sudhir-alifsemi (commented):
rebase and push for automation

sudhir-alifsemi (approved):

PR #125, opened 2025-05-12
drivers: ns16550: Async UART support for PL330 DMA

petrih-alifsemi:
Add support for PL330 DMA to ns16550 UART driver driver to get async UART working properly.

### DIFF
--- a/drivers/dma/dma_pl330.c
+++ b/drivers/dma/dma_pl330.c
@@ -343,6 +343,8 @@ static int dma_pl330_setup_ch(const struct device *dev,
 	sys_write8(OP_DMA_END, dma_exec_addr + offset + 2);
 	sys_write8(OP_DMA_END, dma_exec_addr + offset + 3);
 
+	channel_cfg->loop_counter0 = loop_counter0;
+
 	return 0;
 }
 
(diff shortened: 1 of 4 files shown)


drivers/dma/dma_pl330.h
```diff
@@ -56,13 +56,20 @@
...
 #define DMAC_PL330_CS0		0x100
+#define DMAC_PL330_RTCHCFG	0x400
 #define DMAC_PL330_DBGSTATUS	0xd00
 #define DMAC_PL330_DBGCMD	0xd04
 #define DMAC_PL330_DBGINST0	0xd08
 #define DMAC_PL330_DBGINST1	0xd0c
 #define DMAC_PL330_CR0		0xE00
 #define DMAC_PL330_CRD		0xE14
 
+#define DMA_SARn(ch) (DMAC_PL330_RTCHCFG + 0x20 * (ch))
```
petrih-alifsemi, line 67: change names `DMA_` -> `DMA_PL330_`.

petrih-alifsemi: fixed.

drivers/dma/dma_pl330.c
```diff
@@ -823,10 +825,61 @@ static int dma_pl330_initialize(const struct device *dev)
...
+	const uint32_t reg_base = dev_cfg->reg_base;
+
+	if (!stat || channel >= dev_cfg->max_dma_channels) {
+		return -EINVAL;
+	}
+
+	stat->busy = atomic_get(&channel_cfg->channel_is_active) != DMA_CHANNEL_IS_FREE;
+	stat->dir = channel_cfg->direction;
+	stat->pending_length =
+		sys_read32(reg_base + DMA_PL330_LC1_n(channel)) * (channel_cfg->loop_counter0 + 1) +
```
VeijoPesonen, line 843: A comment which explains the logic wouldn't hurt.

drivers/serial/uart_ns16550.c
```diff
@@ -721,6 +732,15 @@ static int uart_ns16550_configure(const struct device *dev,
 		}
 	}
 
+#if CONFIG_UART_ASYNC_API
+	if (dev_data->async.tx_dma_params.dma_dev != NULL) {
+		struct dma_config *p_dma_cfg =
+			&dev_data->async.tx_dma_params.dma_cfg;
+		p_dma_cfg->source_burst_length = p_dma_cfg->dest_burst_length =
```
VeijoPesonen, line 739: A comment wouldn't hurt.

drivers/serial/uart_ns16550.c
```diff
@@ -1721,16 +1760,34 @@ static int uart_ns16550_rx_enable(const struct device *dev, uint8_t *buf, const
...
+	}
+	uint32_t fcr_reg = FCR_FIFO;
+#if CONFIG_DMA_PL330
+	/*
+	 * Program FIFO: enabled
+	 *   RX generates the interrupt when 1 byte in FIFO
+	 *   TX generates the interrupt when FIFO is 1/4 full
+	 *   Clear TX and RX FIFO if system is not resuming
+	 */
+	fcr_reg += FCR_TX_FIFO_QUARTER | FCR_FIFO_1 | FCR_RCVRCLR | FCR_XMITCLR;
```
VeijoPesonen, line 1785: Do you actually mean this?
```suggestion
	fcr_reg |= FCR_TX_FIFO_QUARTER | FCR_FIFO_1 | FCR_RCVRCLR | FCR_XMITCLR;
```

VeijoPesonen (changes requested):

drivers/serial/uart_ns16550.c
petrih-alifsemi: no, `+=` is more robust way to handle that because values could be different in size.

drivers/dma/dma_pl330.c
petrih-alifsemi: done.

drivers/serial/uart_ns16550.c
petrih-alifsemi: done

drivers/serial/uart_ns16550.c
AkiMakkonen-AlifSemi: Integer promotion happens for bitwise OR so I'm not convinced about the added robustness.
One could also argue that using addition can be less robust... if the fcr_reg had one of the flags (which we are now setting) already set earlier in the code then bitwise OR would give correct result while addition fails.

drivers/serial/uart_ns16550.c
```diff
@@ -721,6 +732,26 @@ static int uart_ns16550_configure(const struct device *dev,
 		}
 	}
 
+#if CONFIG_UART_ASYNC_API
```
tero-alifsemi, line 735: Please use IS_ENABLED( ) check for all config macros

AnttiKauppila-AlifSemi (approved):

drivers/serial/uart_ns16550.c
petrih-alifsemi: hmm, why? Complete useless for preprocessor. I noticed that this file has mixed styles... unfortunately. There are also for example these:
`#if (IS_ENABLED(CONFIG_UART_NS16550_INTEL_LPSS_DMA))`
or
`#if defined(CONFIG_UART_NS16550_INTEL_LPSS_DMA)`
or
`#ifdef CONFIG_UART_NS16550_WA_ISR_REENABLE_INTERRUPT`
or 
`#if UART_NS16550_RESET_ENABLED`

but generally Zephyr code uses basic preprocessor handling with just like this:
`#if CONFIG_UART_INTERRUPT_DRIVEN`.

petrih-alifsemi: all those should be aligned but that would be an another PR, imo

drivers/serial/uart_ns16550.c
petrih-alifsemi: yeah, in that sense I agree. I can modify this even it does not really matter which way this is done. I recall some cases where GCC messed up flag calculation of different types but that could be enhanced nowadays :)
Will be changed...

petrih-alifsemi: done

VeijoPesonen (approved):

tero-alifsemi (approved):

PR #126, opened 2025-05-12
Zrtss 455 add clock ctrl to uart ns16550 v1

Prabhakar-Alifsemi:


sudhir-alifsemi (changes requested):
The oneliner and commit description has to be modified as per the patch. It has nothing to 
do with Alif.

Prabhakar-Alifsemi:
> The oneliner and commit description has to be modified as per the patch. It has nothing to do with Alif.

Updated the oneliner pf the patch

drivers/serial/uart_ns16550.c
```diff
@@ -620,6 +620,11 @@ static int uart_ns16550_configure(const struct device *dev,
 			goto out;
 		}
 
+		ret = clock_control_configure(dev_cfg->clock_dev, dev_cfg->clock_subsys, NULL);
+		if (ret != 0) {
```
sudhir-alifsemi, line 624: if ((ret != 0) && (ret != -ENOSYS)) {
			goto out;
		}

sudhir-alifsemi (changes requested):

drivers/serial/Kconfig.ns16550
```diff
@@ -61,6 +61,15 @@ config UART_NS16550_VARIANT_NS16950
 	  This enables support for 128-bytes FIFO and automatic hardware flow control.
 endchoice
 
+config UART_NS16550_FIFO_SIZE
```
parthitce, line 64: why is this both Kconfig and dts? IMO this is IP's hardware property, so with dts is enough.

parthitce (changes requested):

drivers/serial/uart_ns16550.c
```diff
@@ -1960,6 +1961,7 @@ static DEVICE_API(uart, uart_ns16550_driver_api) = {
 			COND_CODE_1(DT_INST_PROP_OR(n, hw_flow_control, 0),          \
 				    (UART_CFG_FLOW_CTRL_RTS_CTS),                    \
 				    (UART_CFG_FLOW_CTRL_NONE)),                      \
+		.fifo_size = DT_INST_PROP_OR(n, fifo_size, 0),                       \
```
mikter, line 1964: instead of 0 default should you use 1 so no need to check in code

PR #127, opened 2025-05-12
manifest: Add HAL alif

Prabhakar-Alifsemi:
Add hal alif to manifest.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -23,6 +23,8 @@ manifest:
       url-base: https://github.com/zephyrproject-rtos
     - name: babblesim
       url-base: https://github.com/BabbleSim
+    - name: alif
+      url-base: https://github.com/alifsemi
 
   group-filter: [-babblesim, -optional]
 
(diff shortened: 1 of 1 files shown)


sudhir-alifsemi (approved):

west.yml
```diff
@@ -23,6 +23,8 @@ manifest:
       url-base: https://github.com/zephyrproject-rtos
     - name: babblesim
       url-base: https://github.com/BabbleSim
+    - name: alif
+      url-base: https://github.com/alifsemi
```
parthitce, line 27: Why the idea in bigger picture? HAL doesn't have anything device or driver specific abstraction. Here is my observation,
- NO driver or device specific details are part of HAL. May be very few is part of HAL
- IP's are from vendors if am not wrong i.e most of from designware and the drivers are native in Zephyr
- there are radio related blobs and headers, which isn't interest of E7 IMO
- Some files in HAL uses zephyr API's like mutex and so on, what's the idea again here? HAL is HAL, why does it depend on Zephyr?

parthitce (changes requested):

PR #129, opened 2025-05-12
drivers: video: Fix video buffer allocation issue

ChandraBhushan-AlifSemi:
- Added a dedicated section in the SRAM0 region exclusively for the video buffer to prevent overwriting existing data.
- Utilized the k_heap_init API in video_common.c to enable heap initialization within the customizable memory region.


drivers/video/video_common.c
```diff
@@ -8,10 +8,9 @@
 #include <zephyr/drivers/video.h>
 
 #if defined(CONFIG_SOC_FAMILY_ENSEMBLE) && !defined(CONFIG_SOC_SERIES_ENSEMBLE_E1C)
-#define FB_MEM_BASE 0x08000000 /* CONFIG_SOC_FAMILY_ENSEMBLE */
-#elif (defined(CONFIG_SOC_FAMILY_ENSEMBLE) && defined(CONFIG_SOC_SERIES_ENSEMBLE_E1C)) ||          \
-	defined(CONFIG_SOC_SERIES_BALLETTO_B1)
-#define FB_MEM_BASE 0x200a0000 /* CONFIG_SOC_SERIES_ENSEMBLE_E1C */
+#define VIDEO_HEAP_SIZE (CONFIG_VIDEO_BUFFER_POOL_SZ_MAX * CONFIG_VIDEO_BUFFER_POOL_NUM_MAX)
+static struct k_heap video_buffer_pool;
+static uint8_t video_buff[VIDEO_HEAP_SIZE] __aligned(8) __attribute__((section(".video_buffer")));
```
sudhir-alifsemi, line 13: Instead of having separate video buffer, i think we can use the same video buffer placed in the linker script in their soc (ie in their specific linker scripts e7/linker.ld) linkers.

Ideally, we should not be using CONFIG_SOC macros in the driver files.

sudhir-alifsemi (changes requested):

PR #132, opened 2025-05-14
Pinctrl 4.x sup

Prabhakar-Alifsemi:
Add pinctrl support for the Alif Semiconductor


jheiskan81 (commented):
PR is saying generic support both this is only for E7?

silesh-alifsemi (commented):
For the second patch, I think "drivers: pinctrl: Add Alif driver" is a better commit message. Also, I see that other vendors are adding the entire driver support in a single commit (driver support, integrating to the build system and soc pinctrl support). Maybe we should do the same?

sudhir-alifsemi (approved):

silesh-alifsemi (approved):

PR #140, opened 2025-05-16
toolchain: gcc: fixes __ramfunc usage with Clang

VeijoPesonen:
With Clang you can't set long_call attribute on function basis. Instead
of converting all calls to long calls let the linker create veneers when
necessary.

Depends on PR alifsemi/hal_alif#34.

Upstream PR: zephyrproject-rtos/zephyr#90148


sudhir-alifsemi (dismissed):

VeijoPesonen:
Manifest updated.

petrih-alifsemi (approved):

sudhir-alifsemi (approved):

PR #152, opened 2025-06-02
drivers: video: hm0360: Optimize the Camera Initialization time

YogenderKumar-AlifSemi:
Add all the writes to camera sensor as burst writes, so that the address is not required to be updated for every sensor register write. Also, remove any un-necessary delays from the driver code.


RupeshKumar-AlifSemi:
Please fix checkpatch issues.

sudhir-alifsemi (approved):

RupeshKumar-AlifSemi (approved):

PR #153, opened 2025-06-04
dts: pinctrl: ensemble : update ospi1_rxds_c pin config

GaneshRamani-AlifSemi:
Update ospi1_rxds_c pin to P10:7 and function 7.


VeijoPesonen (approved):
I would add the information - how it was wrong earlier - on the commit message but no need to do it if nobody else has a change request.

GaneshRamani-AlifSemi:
> I would add the information - how it was wrong earlier - on the commit message but no need to do it if nobody else has a change request.

it was not updated after B0. The PIN10.7 declared from this patch.

PR #159, opened 2025-06-12
drivers: video: mt9m114 csi support

YogenderKumar-AlifSemi:
Added support for MT9M114 CSI-2 interface. Moved clock support for CSI-2 to Controller driver and removed clock enablement for CSI-2 controller from soc files.


YogenderKumar-AlifSemi:
> we should use SE services to poke the CGU registers. For now, I prefer to enable the CGU clocks required for this from the application using SE services call.

Should the call to SE be made from the driver while it needs that clock, or should that call be made from main.c before running of the application. When I tested setting the 100MHz clock during set_format call, the application still worked; so either of the above two methods (calling from driver or application) should work

VeijoPesonen:
Please updates SDK's west manifest.

VeijoPesonen:
Update the PR topic to be in line with all the other PRs.

sudhir-alifsemi:
> > we should use SE services to poke the CGU registers. For now, I prefer to enable the CGU clocks required for this from the application using SE services call.
> 
> Should the call to SE be made from the driver while it needs that clock, or should that call be made from main.c before running of the application. When I tested setting the 100MHz clock during set_format call, the application still worked; so either of the above two methods (calling from driver or application) should work.
> 
> Edit: the clock enable is not available in SE.

Yes, the clock enable API has to be added to the se_services. Ideally these SE apis has to be invoked from the clock control driver.

drivers/video/video_csi_dw.c
```diff
@@ -506,13 +519,81 @@ static const struct video_driver_api csi2_dw_driver_api = {
...
+	if (ret) {
+		LOG_ERR("Pixel-CLK source configuration failed! ret - %d", ret);
+		return ret;
+	}
+
+	/* Set maximum rate for IPI interface of CSI and CPI. This is
+	 * reconfigured during configuration time of CSI and D-PHY.
+	 */
+	ret = clock_control_set_rate(config->clk_dev, config->pix_cid,
+			(clock_control_subsys_rate_t) MHZ(200));
```
ssekar15, line 563: get rate from dts

drivers/video/video_csi_dw.c
```diff
@@ -506,13 +519,81 @@ static const struct video_driver_api csi2_dw_driver_api = {
...
 };
 
+#if DT_ANY_INST_HAS_PROP_STATUS_OKAY(clocks)
+static int csi2_dw_enable_clocks(const struct device *dev)
+{
+	const struct csi2_dw_config *config = dev->config;
+	int ret;
+
+	/* Enable CSI clock. */
+	ret = clock_control_on(config->clk_dev, config->csi_cid);
```
ssekar15, line 529: check the clk_dev is valid before proceeding

PR #165, opened 2025-06-17
alif: ensemble: spi3 pinmux update

GaneshRamani-AlifSemi:
SPI3 and OSPI1_RxDS pin updated.


GaneshRamani-AlifSemi:
> Is there a way you can generate the pinctrl dtsi based on input file or so? Otherwise it will always be preparing the who file for all the pins manually.

i don't think there is tool for this to generate with us. it is updated manually so far.

parthitce:
> > Is there a way you can generate the pinctrl dtsi based on input file or so? Otherwise it will always be preparing the who file for all the pins manually.
> 
> i don't think there is tool for this to generate with us. it is updated manually so far.

Doesn't scale in longer run IMO. E7, E8, E9 and what not is upcoming.

parthitce (approved):
ok, this is going to main branch which is downstream fork for alif. if this is already tested, am fine with it.

VeijoPesonen:
@GaneshRamani-AlifSemi Rebase and reopen if you plan to continue working with this.

PR #168, opened 2025-06-27
drivers: add dave2d driver

YevhenBobovskyi-AlifSemi:
Add D/AVE2D driver with hardware abstraction layer

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -128,6 +128,11 @@ manifest:
     - name: cmsis-nn
       revision: 0c8669d81381ccf3b1a01d699f3b68b50134a99f
       path: modules/lib/cmsis-nn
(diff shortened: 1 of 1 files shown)


parthitce:
@YevhenBobovskyi-AlifSemi I have few questions regarding this driver,
1. Is the whole IP / driver written from scratch from Alif? or is it also part from TES?
2. How to use this IP with Alif board?
3. With Zephyr we have the video driver structure to expose the usage. As with this as 2D graphics accelerator, are there any API grouping done?
4. Do we have sample application to test this driver?

jheiskan81:
PR which include 1 commit and touch 239 files is too big to review.

PR #171, opened 2025-07-02
[fix] alif_flash was not working when the build target was in subfolder

Szendvics:
[fix] alif_flash was not working when the build target was in subfolder

alif_flash runner was broken when a build-dir was set:
```
./build/alif_e7_dk_rtss_he/blinky
```

```
west flash --runner alif_flash --build-dir ./build/alif_e7_dk_rtss_he/blinky
```

that was caused by the too early cd and by the hardwired location of the zephyr.bin
also restoring the directory was missing after the ToC was done/failed


GaneshRamani-AlifSemi (approved):
Thanks !
External build directory configuration not considered so far!

RupeshKumar-AlifSemi (approved):

rajranjan-alifsemi (approved):

PR #178, opened 2025-07-23
Manifest update for HCI driver

jheiskan81:
Balletto support DMA event router for HCI uart.

Depends on alifsemi/hal_alif#48.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -139,7 +139,7 @@ manifest:
       groups:
         - fs
     - name: hal_alif
-      revision: fc63076a35e235c2e4ea24361f04774c60bbb682
+      revision: 0db33a7136cc1512b292a4dbbd0e086d0e8a888e
       path: modules/hal/alif
       remote: alif
       groups:


drivers/serial/uart_ns16550.c
```diff
@@ -572,6 +572,52 @@ static void set_baud_rate(const struct device *dev, uint32_t baud_rate, uint32_t
...
+ *
+ * This function configures the event router to connect UART peripheral
+ * to the PL330 DMA controller.
+ *
+ * @param dma_group DMA group (0-3)
+ * @param dma_request DMA request number (0-31)
+ *
+ * @return 0 on success, negative errno on failure
+ */
+static int configure_dma_event_router(const uint32_t dma_group, const uint32_t dma_request)
```
petrih-alifsemi, line 596: @sudhir-alifsemi do you have some idea related to `EVTRTR` configuration? or is it ok to configure this here for now?
I know that this has been done in "user layer" before...

petrih-alifsemi: I know, is there any plan when that would be available?

petrih-alifsemi (approved):

jheiskan81:
Manifest point now to real commit hash.

petrih-alifsemi (approved):

PR #184, opened 2025-08-19
drivers: mipi_dsi: Fix payload read

tero-alifsemi:
Fix register checks in dsi_dw_read_payload


VeijoPesonen (changes requested):
Please define what you're actually reading, DSI_CMD_PKT_STATUS_OK?

VeijoPesonen:
And update the SDK west manifest.

VeijoPesonen (approved):

PR #188, opened 2025-08-25
scripts: west_commands: runners: add debug support for Alif SoC

GaneshRamani-AlifSemi:
Add debug support using the J-Link adapter in the alif_flash runner.

The J-Link runner is not used for Alif SoCs since flashing to MRAM requires Alif Tools. The alif_flash runner integrates both flashing (via Alif Tools) and debugging (via J-Link)

### DIFF
--- a/boards/arm/alif_b1_dk/board.cmake
+++ b/boards/arm/alif_b1_dk/board.cmake
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
+board_runner_args(alif_flash "--device=AB1C1F4M51820" "--speed=4000")
+
 include(${ZEPHYR_BASE}/boards/common/alif_flash.board.cmake)
 
--- a/scripts/west_commands/runners/alif_flash.py
+++ b/scripts/west_commands/runners/alif_flash.py
@@ -4,13 +4,17 @@
 Runner for Alif binary image burner.
 """
 import os
+import sys
 import json
 import shutil
+import re
+import fdt
 
 from pathlib import Path
-from runners.core import ZephyrBinaryRunner, RunnerCaps
+from runners.core import ZephyrBinaryRunner, RunnerCaps, FileType
 
-import fdt
+DEFAULT_JLINK_GDB_PORT = 2331
+DEFAULT_JLINK_GDB_SERVER = 'JLinkGDBServerCL' if sys.platform == 'win32' else 'JLinkGDBServer'
 
 class AlifImageBinaryRunner(ZephyrBinaryRunner):
     '''Runner front-end for Alif Image Flasher.'''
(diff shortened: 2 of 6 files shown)


boards/arm/alif_b1_dk/board.cmake
```diff
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
+board_runner_args(alif_flash "--device=AB1C1F4M51820" "--speed=4000")
```
KimmoVaisanen-AlifSemi, line 3: Could we use higher speeds? I know Segger defaults to 4MHz, but I think our HWs work fine at least up to 20MHz (probably even higher).

GaneshRamani-AlifSemi: @KimmoVaisanen-AlifSemi , shall we keep `--speed=auto` so not required to hardcode any value ?

KimmoVaisanen-AlifSemi: @KarolSaja-AlifSemi do you have opinion on this? Our VSCode template uses speed 15000. For me auto sounds good, if it works reliably.

KarolSaja-AlifSemi: If it writes MRAM over serial interface, why do we need to specify anything J-Link related?

FYI, J-Link can also write the TOC as it lives in the user area of the MRAM. The only issue is that you need to be able to connect to the core already in order to do that (unless SEGGER adds AXI support eventually).

KimmoVaisanen-AlifSemi (approved):

KarolSaja-AlifSemi (dismissed):
Looks good, though if python release of SE tools is used, specific routines can be invoked directly as python functions rather than executables (which results in two python runtimes being set up to run this task)

GaneshRamani-AlifSemi:
added co-pilot suggestion !

scripts/west_commands/runners/alif_flash.py
```diff
@@ -99,6 +140,104 @@ def do_run(self, command, **kwargs):
             os.chdir(old_cwd)
             self.logger.info(f"Returned to working directory {old_cwd}")
 
+    def debug(self, attach = False, **kwargs):
+            '''debug '''
+            #check tool availability
+            self.require(self.gdbserver)
+
```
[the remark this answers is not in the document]
GaneshRamani-AlifSemi: added suggested change !

scripts/west_commands/runners/alif_flash.py
```diff
@@ -99,6 +140,104 @@ def do_run(self, command, **kwargs):
...
+                client_cmd += ['-ex', 'monitor reset']
+
+            #launch Server and Client subprocess
+            self.run_server_and_client(server_cmd, client_cmd)
+
+    def do_run(self, command, **kwargs):
+
+        #Attach : Launch debug session to the running process.
+        if command == 'attach':
+            self.debug(attach = True, **kwargs)
```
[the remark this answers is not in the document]
GaneshRamani-AlifSemi: done!

Szendvics:
Thank you for the PR however I don't see why it should launch a JLINK debugger here.
When I popped up the issue the main goal was to have a crosscheck within alif_flash if the Alif SE Tools setup is aligned to the west target board and prevent flashing the E7 boards as E8.

scripts/west_commands/runners/alif_flash.py
```diff
@@ -24,38 +28,72 @@ class AlifImageBinaryRunner(ZephyrBinaryRunner):
...
+        self.iface = iface
+        self.speed = speed
         self.erase = bool(erase)
         self.reset = bool(reset)
+        self.gdb_cmd = [cfg.gdb] if cfg.gdb else None
+        self.file = cfg.file
+        self.file_type = cfg.file_type
+        self.hex_name = cfg.hex_file
+        self.elf_name = cfg.elf_file
+        self.tui_arg = ['-tui'] if tui else []
```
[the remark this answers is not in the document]
GaneshRamani-AlifSemi: fixed !

PR #195, opened 2025-09-22
soc: arm: alif: disable CONFIG_INIT_ARCH_HW_AT_BOOT

Tanay-AlifSemi:
- Disable CONFIG_INIT_ARCH_HW_AT_BOOT to avoid redundant cache handling.
- Cache is already enabled from SystemInit()(z_arm_platform_init).
- Remove cache enable from balletto_b1_dk_rtss_he_init(), as it is no longer needed (previously required only because CONFIG_INIT_ARCH_HW_AT_BOOT disabled the cache).


jheiskan81 (dismissed):
looks good for me

jheiskan81 (approved):

PR #202, opened 2025-10-07
Update README.rst

CijaSathish-AlifSemi:
Tweaks in the README

### DIFF
--- a/samples/sensor/icm42670/README.rst
+++ b/samples/sensor/icm42670/README.rst
@@ -14,7 +14,7 @@ startup.
 Wiring
 *******
 
-This sample uses an external breakout for the sensor.  A devicetree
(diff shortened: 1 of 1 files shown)


samples/sensor/icm42670/README.rst
```diff
@@ -14,7 +14,7 @@ startup.
 Wiring
 *******
 
-This sample uses an external breakout for the sensor.  A devicetree
+No wiring needed as this is an on-board sensor. A devicetree
```
KimmoVaisanen-AlifSemi, line 17: This is a generic Zephyr example, so this is maybe right for Alif devices but not all devices. So, this needs some different wording.

KimmoVaisanen-AlifSemi (changes requested):

PR #352, opened 2026-01-30
sd: Remove multiblock check

Deepak-AlifSemi:
CMD23 check for multiblock transfer need to handle in
the low level sdhc driver, so removed from subsys.

### DIFF
--- a/subsys/sd/sdmmc.c
+++ b/subsys/sd/sdmmc.c
@@ -207,24 +207,6 @@ static int sdmmc_read_scr(struct sd_card *card)
 	return 0;
 }
 
-/* Sets block count for SD card */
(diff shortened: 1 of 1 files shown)


sudhir-alifsemi (commented):
Typo in the git description

sudhir-alifsemi (changes requested):
subsys: sd: check for CMD23 flag

sudhir-alifsemi (changes requested):

sudhir-alifsemi (changes requested):
Please update the git subject line according to the patch. Also, the PR subject line too.

sudhir-alifsemi (approved):

RupeshKumar-AlifSemi (approved):

PR #387, opened 2026-02-25
dts: arm: alif: E1C NS region aligned wrt TGU size

Deepak-AlifSemi:
E1C Non-Secure region size aligned wrt. TGU block size for the ensemble E1C.


sudhir-alifsemi (approved):

sudhir-alifsemi (commented):
Isn't the address 20020000 aligned to 128KB ?

PR #418, opened 2026-03-12
dts: arm: alif: Add sd reset line

Deepak-AlifSemi:
Added SD reset lines for Ensemble E1C.
Added pinmux and increased drive strength


RupeshKumar-AlifSemi (approved):

sudhir-alifsemi (changes requested):
There should not be any mention of internal names. Also the sd reset line is for the board, not soc. Please update the git description accordingly

silesh-alifsemi (changes requested):

silesh-alifsemi (approved):

sudhir-alifsemi (approved):

PR #437, opened 2026-03-25
mpu table updates for MRAM storage for mcuboot support

silesh-alifsemi:
As the writeable regions of the MRAM need to be configured with device memory attributes, update the MPU table for ensemble and balletto to support mcuboot use cases. This basically partitions the MRAM storage area into either an executable flash region or a writeable region with device memory attributes.

Following three scenarios need to be handled based on the requirements of the image (bootloader, mcuboot compatible app, regular app).

1. mcuboot bootloader builds:
   executable flash area: boot partition only
   writeable device area: slot0, slot1, scratch, and storage partitions

2. mcuboot-compatible application builds:
   executable flash area: boot + slot0 (minus last sector for trailer)
   writable device area: slot0 trailer sector + slot1 + scratch + storage partitions.

3. regular non-mcuboot application builds(the existing setup):
   executable flash area: boot + slot0 + slot1 + scratch
   writable device area: storage partition only

### DIFF
--- a/soc/alif/balletto/common/mpu_regions.c
+++ b/soc/alif/balletto/common/mpu_regions.c
@@ -24,40 +24,87 @@
 	.r_limit = REGION_LIMIT_ADDR(base, size),  \
 }
 
+#define MRAM_SECTOR_SIZE		DT_PROP(DT_NODELABEL(mram_storage), erase_block_size)
+
 #define MRAM_BOOT_PARTITION_ADDR	DT_FIXED_PARTITION_ADDR(DT_NODELABEL(boot_partition))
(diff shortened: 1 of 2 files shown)


jheiskan81 (approved):
Thanks for fixing this.

KimmoVaisanen-AlifSemi (approved):
I think this is ok for now until more flexible solution is available.

This is fully tied to current MRAM memory layout. If user modifies the lyout, this needs to be modified also, so I think some sort of note should be added to DTS where MRAM layout is defined.

sudhir-alifsemi (approved):

PR #438, opened 2026-03-25
drivers: i2c: Correct i2c_dw target implementation

Shreehari-AlifSemi:
The I2C target implementation for i2c_dw does not follow the zephyr target specification. The following three issues are corrected:

1) The read_requested CB is called on every byte of the transaction
    - it must be called only on the first byte of the transaction
2) The read_processed CB is called on every byte of the transaction
    - it must be called only on bytes following the first byte
3) Bytes returned in the data ptr param of read_processed are not sent to the controller
     -These bytes must be sent to the controller

### DIFF
--- a/drivers/i2c/i2c_dw.c
+++ b/drivers/i2c/i2c_dw.c
@@ -541,12 +541,17 @@ static void i2c_dw_isr(const struct device *port)
 				}
 			}
 #else
-			if (slave_cb->read_requested) {
-				slave_cb->read_requested(dw->slave_cfg, &data);
-				i2c_dw_write_byte_non_blocking(port, data);
-			}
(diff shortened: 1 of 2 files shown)


silesh-alifsemi (changes requested):
As this is an upstream patch, please use git-cherry-pick so that the details are preserved.

Shreehari-AlifSemi:
> As this is an upstream patch, please use git-cherry-pick so that the details are preserved.

I did it, but there was conflict during the process. So, I fixed and pushed it. I just added one statement of mine in it

silesh-alifsemi (approved):

RupeshKumar-AlifSemi (approved):

PR #439, opened 2026-03-25
drivers: sdhc: Add synopsys dwc sdhc driver

Deepak-AlifSemi:
Added initial version of synopsys dwc sdhc driver.
Supports 3.3v and 1.8v operation.
Supports SD/SDIO modes at 1, 4 bit bus, 400khz, 25mhz, 50mhz speed.


drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1152 @@
...
+#if defined(CONFIG_SDHC_DWC_ADMA)
+	uint32_t remaining = total_len;
+	uint32_t offset = 0;
+	uint32_t desc_num = 0;
+
+	while (remaining > 0 && desc_num < CONFIG_SDHC_DWC_ADMA_MAX_DESC) {
+		uint32_t chunk = (remaining > DWC_SDHC_ADMA2_DESC_MAX_LEN)
+				 ? DWC_SDHC_ADMA2_DESC_MAX_LEN : remaining;
+		uint32_t addr = SDHC_DMA_ADDR((const volatile void *)
+						((uint32_t)sdhc_data->data + offset));
```
silesh-alifsemi:
```suggestion
						((uintptr_t)sdhc_data->data + offset));
```

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1152 @@
...
+	const struct sdhc_dwc_config *config = dev->config;
+	struct dwc_sdhc_regs *regs = config->regs;
+	uint32_t reset_timeout = DWC_SDHC_SW_RST_TIMEOUT;
+
+	regs->DWC_SDHC_SW_RST_R = reset_mask;
+	while (regs->DWC_SDHC_SW_RST_R != 0 && reset_timeout--) {
+		k_busy_wait(1);
+	}
+
+	if (reset_timeout == 0) {
```
silesh-alifsemi: On a timeout, `reset_timeout` will never be 0 because of the post decrement above ?

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1152 @@
...
+				struct sdhc_command *cmd, bool data_present)
+{
+	const struct sdhc_dwc_config *config = dev->config;
+	struct dwc_sdhc_regs *regs = config->regs;
+	uint32_t cmd_timeout = cmd->timeout_ms * USEC_PER_MSEC;
+
+	/* Check CMD line idle */
+	while (regs->DWC_SDHC_PSTATE_REG & DWC_SDHC_CMD_INHIBIT_Msk) {
+		if (!cmd_timeout--) {
+			return true;
```
silesh-alifsemi: As per the function name, we should return` true` in the case the cmd/data lines are _not busy_.

drivers/sdhc/sdhc_dwc.h
```diff
@@ -0,0 +1,302 @@
...
+#define DWC_SDHC_INTERNAL_CLK_STABLE_Msk     0x2U
+#define DWC_SDHC_CLK_EN_Msk                  0x4U
+#define DWC_SDHC_PLL_EN_Msk                  0x8U
+#define DWC_SDHC_CLK_GEN_SEL_Pos             5U
+#define DWC_SDHC_DIV_CLK_MODE                0x0U
+#define DWC_SDHC_CLK_GEN_SEL_Msk             (DWC_SDHC_DIV_CLK_MODE << DWC_SDHC_CLK_GEN_SEL_Pos)
+#define DWC_SDHC_UPPER_FREQ_SEL_Pos          6U
+#define DWC_SDHC_FREQ_SEL_Pos                8U
+#define DWC_SDHC_BASE_CLK_FREQ_Pos           7U
+#define DWC_SDHC_BASE_CLK_FREQ_Msk           (0xFFU << DWC_SDHC_FREQ_SEL_Pos)
```
silesh-alifsemi, line 191: Should the shift be `DWC_SDHC_BASE_CLK_FREQ_Pos`  ?

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1152 @@
+/*
+ * Copyright (c) 2026 Alif Semiconductor
```
silesh-alifsemi:
```suggestion
 * SPDX-FileCopyrightText: Copyright Alif Semiconductor
```

drivers/sdhc/Kconfig.dwc
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) 2026 Alif Semiconductor
```
silesh-alifsemi:
```suggestion
# SPDX-FileCopyrightText: Copyright Alif Semiconductor
```

dts/bindings/sdhc/snps,dwc-sdhc.yaml
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2024 Alif Semiconductor
```
silesh-alifsemi:
```suggestion
# SPDX-FileCopyrightText: Copyright Alif Semiconductor
```

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1152 @@
...
+
+		/* Deassert reset */
+		ret = gpio_pin_set_dt(&config->reset_gpio, 0);
+		if (ret < 0) {
+			return ret;
+		}
+	}
+
+	/* Software reset all */
+	sdhc_dwc_reset(dev);
```
silesh-alifsemi: check the return value?

silesh-alifsemi (changes requested):

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1173 @@
...
+			return ret;
+		}
+
+		/* Assert reset */
+		ret = gpio_pin_set_dt(&config->reset_gpio, 1);
+		if (ret < 0) {
+			return ret;
+		}
+
+		k_msleep(100);
```
silesh-alifsemi, line 1074: Do we need to deassert the gpio line?

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1173 @@
...
+		}
+	}
+
+	/* Error Interrupts */
+	if (eis) {
+		LOG_DBG("ERR_INT_STAT: 0x%04x", eis);
+		sdhc_dwc_dump_regs(dev);
+		k_event_post(&data->irq_event, ERR_INTR_STATUS_EVENT(eis));
+
+		sdhc_dwc_hw_reset(dev, DWC_SDHC_SW_RST_CMD_Msk | DWC_SDHC_SW_RST_DAT_Msk);
```
silesh-alifsemi: Should we really call `sdhc_dwc_hw_reset` (which has a busy wait for reset timeout) from an ISR?

drivers/sdhc/sdhc_dwc.c
```diff
@@ -20,8 +20,14 @@
 
 #include "sdhc_dwc.h"
 
+#ifdef CONFIG_SDHC_DWC_DMA_ADDR_TRANSLATE
+#if defined(CONFIG_SOC_FAMILY_ENSEMBLE) || defined(CONFIG_SOC_FAMILY_BALLETTO)
```
silesh-alifsemi, line 24: It would be better if we can avoid  `SOC_FAMILY_ENSEMBLE/SOC_FAMILY_BALLETTO` to make the driver more generic. Could we do a weak default implementation (with no translation) for `SDHC_DMA_ADDR` that platforms can then override with whatever they want (for eg. alif SoCs can override with` local_to_global`)?

drivers/sdhc/Kconfig.dwc
```diff
@@ -0,0 +1,50 @@
...
+
+if SDHC_DWC
+
+config SDHC_BUFFER_ALIGNMENT
+	int "SDHC buffer alignment"
+	default 512
+
+config SDHC_DESCRIPTOR_SECTION
+	string
+	default ".alif.ns"
```
silesh-alifsemi: `.alif_ns` or maybe we should mention the default section that the descriptors go to here(without this override)? Mentioning an alif specific linker section might not be good for a generic driver.

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1173 @@
...
+{
+	return sdhc_dwc_hw_reset(dev, DWC_SDHC_SW_RST_ALL_Msk);
+}
+
+static int sdhc_dwc_dma_init(struct dwc_sdhc_regs *regs, struct sdhc_dwc_data *data,
+			      struct sdhc_data *sdhc_data, bool read)
+{
+	uint32_t total_len = sdhc_data->blocks * sdhc_data->block_size;
+
+	if (IS_ENABLED(CONFIG_DCACHE) && !read) {
```
silesh-alifsemi: Please use the guard consistently (and correctly), the cache maintenance operations are guarded below by `CACHE_MANAGEMENT` (different from here).

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1173 @@
...
+	LOG_DBG("Voltage set to %s (pwr_ctrl=0x%02x)",
+		voltage == SD_VOL_3_3_V ? "3.3V" :
+		voltage == SD_VOL_3_0_V ? "3.0V" :
+		voltage == SD_VOL_1_8_V ? "1.8V" : "unknown",
+		regs->DWC_SDHC_PWR_CTRL_R);
+
+	return 0;
+}
+
+static int sdhc_dwc_set_power(struct dwc_sdhc_regs *regs, enum sdhc_power state)
```
silesh-alifsemi: Should this be `static void` ?

drivers/sdhc/sdhc_dwc.c
```diff
@@ -0,0 +1,1173 @@
...
+	struct dwc_sdhc_regs *regs = config->regs;
+	uint32_t pstate, dat_line_status;
+
+	uint32_t reg = regs->DWC_SDHC_PSTATE_REG;
+
+	pstate = reg & (DWC_SDHC_DAT_INHIBIT_Msk | DWC_SDHC_CMD_INHIBIT_Msk);
+	dat_line_status = reg & DWC_SDHC_CMD_DATA_LINE_STATUS_Msk;
+
+	if (pstate) {
+		return true;        /* CMD or DAT inhibit set -> busy */
```
silesh-alifsemi, line 836: Should the prototype be changed to `static bool` ?

silesh-alifsemi (changes requested):

drivers/sdhc/sdhc_dwc.c
Deepak-AlifSemi: The Zephyr SDHC API defines card_busy as int (*card_busy)(const struct device *dev) returning int (0/1/-EIO). Changing the implementation to bool creates a type mismatch with the API's function pointer.

drivers/sdhc/sdhc_dwc.c
Deepak-AlifSemi: Weak default implementation in the driver is not as per the standard, so keeping this will create an issue for upstreaming.

silesh-alifsemi (approved):

RupeshKumar-AlifSemi (approved):

PR #442, opened 2026-03-26
drivers: dma: pl330: add scatter-gather block chaining support

CijaSathish-AlifSemi:
Implement scatter-gather support in the PL330 DMA driver by walking the dma_block_config linked list. Each block is transferred sequentially via the ISR without CPU involvement between blocks.

### DIFF
--- a/drivers/dma/dma_pl330.c
+++ b/drivers/dma/dma_pl330.c
@@ -30,6 +30,45 @@ LOG_MODULE_REGISTER(dma_pl330);
 static int dma_pl330_submit(const struct device *dev, uint64_t dst,
 			    uint64_t src, uint32_t channel, uint32_t size);
 
+
+/*
+ * Start a scatter-gather block transfer
+ *
+ * Loads the given block's configuration into the channel and initiates
+ * the DMA transfer for that block. Used for both initial block start and
+ * advancing through the scatter-gather chain.
+ */
+static int dma_pl330_start_block(const struct device *dev,
+				 uint32_t channel,
+				 struct dma_block_config *blk)
+{
+	const struct dma_pl330_dev_data *const dev_data = dev->data;
+	struct dma_pl330_ch_config *channel_cfg = &dev_data->channels[channel];
+
+	/* ensure there is a valid block to transfer */
+	if (!blk) {
+		return -EINVAL;
+	}
+
+	/* load the current block's source, destination and size
+	 * into the channel config so dma_pl330_submit() can use them
+	 */
+	channel_cfg->src_addr = local_to_global(UINT_TO_POINTER(blk->source_address));
+	channel_cfg->dst_addr = local_to_global(UINT_TO_POINTER(blk->dest_address));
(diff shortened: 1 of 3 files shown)


drivers/dma/dma_pl330.h
```diff
@@ -197,6 +197,19 @@ struct dma_pl330_ch_config {
...
+	 * Scatter-gather support fields
+	 *
+	 * head_block: Pointer to the first block in the scatter-gather chain
+	 * current_block: Pointer to the block currently being transferred
+	 * current_block_remain: Bytes remaining in the current block
+	 * block_count: Total number of blocks in the scatter-gather chain
+	 */
+	struct dma_block_config *head_block;
+	struct dma_block_config *current_block;
+	size_t current_block_remain;
```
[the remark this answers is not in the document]
jheiskan81: Should this feature have a kConfig for enable now it will be standard behavior. I just thinking upstream point of view because this is generic driver.

jheiskan81 (changes requested):

jheiskan81 (changes requested):
Please add kConfig for this new feature.

sudhir-alifsemi (approved):

jheiskan81 (approved):

drivers/dma/dma_pl330.h
```diff
@@ -197,6 +197,19 @@ struct dma_pl330_ch_config {
...
+	 * Scatter-gather support fields
+	 *
+	 * head_block: Pointer to the first block in the scatter-gather chain
+	 * current_block: Pointer to the block currently being transferred
+	 * current_block_size: Size in bytes of the current block
+	 * block_count: Total number of blocks in the scatter-gather chain
+	 */
+	struct dma_block_config *head_block;
+	struct dma_block_config *current_block;
+	size_t current_block_size;
```
silesh-alifsemi: where is this field used (other than the writes into it, it is never read from)?

drivers/dma/dma_pl330.c
```diff
@@ -692,6 +728,25 @@ static int dma_pl330_configure(const struct device *dev, uint32_t channel,
...
+	channel_cfg->head_block = cfg->head_block;
+	channel_cfg->current_block = cfg->head_block;
+	channel_cfg->current_block_size = cfg->head_block->block_size;
+
+	channel_cfg->block_count = 0;
+	for (struct dma_block_config *blk = cfg->head_block; blk != NULL; blk = blk->next_block) {
+		channel_cfg->block_count++;
+	}
+
+	if (channel_cfg->block_count == 0) {
```
silesh-alifsemi: If we are using `block_count` and the above loop to verify that at least one block is available, a simple check for a non-NULL `cfg->head_block` will be enough and we can get rid of `block_count` and the above loop.

silesh-alifsemi (changes requested):

silesh-alifsemi (approved):

PR #453, opened 2026-04-01
soc: alif: add HE_DMA_SEL LPSPI mux config to soc_init

sudhir-alifsemi:
Configure the HE_DMA_SEL register for LPSPI during SoC initialization for both ensemble and balletto families. The DMA controller selection is derived at build time from the DTS dmas property of the LP-SPI node.

E1C and B1 always use DMA2, with the register value determined by the DMA group. E3/E5/E7/E4/E6/E8 support both DMA2 and DMA0, detected via the event router phandle in the dmas specifier.

### DIFF
--- a/soc/alif/ensemble/common/soc_common.c
+++ b/soc/alif/ensemble/common/soc_common.c
@@ -7,6 +7,7 @@
 #include <zephyr/arch/cpu.h>
 #include <soc_common.h>
 #include <zephyr/pm/policy.h>
+#include <zephyr/dt-bindings/dma/alif_dma_event_router.h>
 
 #if CONFIG_ENSEMBLE_GEN2 /* ENSEMBLE_GEN2 SoC */
 /* GPIO: enable debounce clock / divisor. */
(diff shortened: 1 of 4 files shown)


soc/alif/ensemble/common/soc_common.c
```diff
@@ -98,6 +99,53 @@ static int soc_pm_unlock_boot_states(void)
...
+ */
+#if IS_ENABLED(CONFIG_RTSS_HE)
+#if IS_ENABLED(CONFIG_SOC_SERIES_E1C)
+#if DT_NODE_HAS_STATUS(DT_NODELABEL(lpspi0), okay) && DT_NODE_HAS_PROP(DT_NODELABEL(lpspi0), dmas)
+static void soc_configure_he_dma_sel_lpspi(void)
+{
+	uint32_t dma_group = ALIF_DMA_DECODE_GROUP(
+		DT_PHA_BY_IDX(DT_NODELABEL(lpspi0), dmas, 0, channel));
+	/* group 1 = 0x0 (DMA2 group 1); group 2+ = 0x1 (DMA2 group 2) */
+	uint32_t sel_val = (dma_group == 1U) ? 0x0U : 0x1U;
```
RupeshKumar-AlifSemi, line 118: can we define macros for this dma groups and use them instead of Hard coded values?

soc/alif/ensemble/common/soc_common.c
```diff
@@ -98,6 +99,53 @@ static int soc_pm_unlock_boot_states(void)
 }
 SYS_INIT(soc_pm_unlock_boot_states, APPLICATION, 0);
 
+/* Configure HE_DMA_SEL register for LP-SPI based on DTS dmas property.
```
RupeshKumar-AlifSemi, line 102: Can this section be generalized to handle other peripherals as well?  LPSPI is Just a specific case?

RupeshKumar-AlifSemi (changes requested):

soc/alif/ensemble/common/soc_common.c
sudhir-alifsemi: The HE_DMA_SEL register field values (0x0, 0x1, 0x2) are hardware-defined encoding specific to each SoC series. 
On E-series(Except E1C): 0x0 means DMA2, 0x1 means DMA0 group 1, 0x2/0x3 means DMA0 group 2.
On B1/E1C, 0x0 means DMA2 group 1 and 0x1/0x2/0x3 mean DMA2 group 2. So macros would require separate sets per SoC series.

So, I think the best way is to compute the value conditionally based on the series and the decoded dma_group from DTS.

soc/alif/ensemble/common/soc_common.c
sudhir-alifsemi: Each LP peripheral (LPUART, LPI2S, LPPDM, LPI3C, LPSPI) has a different bit field in HE_DMA_SEL, has a different DTS node label, and may only exist on certain SoC series. Whether a peripheral uses DMA is application-defined using the DTS property.
A generic handler would need per-peripheral bit positions, node labels, and per-series existence checks — effectively reproducing separate functions inside one wrapper.

So, the current approach of one focused function per peripheral (to be added as other LP peripherals uses DMA) is cleaner.

soc/alif/ensemble/common/soc_common.c
RupeshKumar-AlifSemi: ok

RupeshKumar-AlifSemi (approved):

silesh-alifsemi (approved):

PR #458, opened 2026-04-02
drivers: video: alif_cam: handle dual endpoint stream control

Shivakumar-AlifSemi:
Skip video_stream_stop on endpoint_dev when ISP endpoint is active to avoid stopping a shared stream. Ignore -EBUSY on stream start as the pipeline may already be running via ISP. Add CAM_CFG register log on stream start for debugging.


silesh-alifsemi (commented):
@Shivakumar-AlifSemi Please have a look at the copilot comments and resolve them.

PR #470, opened 2026-04-17
drivers: flash: allow read/write for odd start address + odd length

Tanay-AlifSemi:
Allow read/write for odd start address + odd length (odd end address) in DDR mode.
For E8 DK board fix write-block-size to 2.

### DIFF
--- a/drivers/flash/flash_ospi_is25wx.c
+++ b/drivers/flash/flash_ospi_is25wx.c
@@ -463,22 +463,143 @@ static int flash_is25wx_ospi_write(const struct device *dev, off_t address, cons
 		return -EINVAL;
 	}
 
-	/*In DDR Mode bytes and length should be in multiple of write_block_size */
-	if (length % f_param->write_block_size
-		|| !(IS_ALIGNED(buffer, f_param->write_block_size))) {
-		return -EINVAL;
-	}
-
 	ret = k_sem_take(&dev_data->sem, K_MSEC(MAX_SEM_TIMEOUT));
 	if (ret != 0) {
 		return ret;
 	}
 
-	/* number of block count */
-	cnt = length / f_param->write_block_size;
 	data_ptr = (uint8_t *) buffer;
 
-	while (cnt) {
+	/* Handle prefix bytes if start address is not word-aligned.
+	 * Write one full block at the aligned-down address, placing user
+	 * data at the correct byte offset within the word.  Padding with
+	 * erase_value (0xFF) is a no-op for flash programming.
+	 */
+	size_t addr_offset = address & (f_param->write_block_size - 1);
+
+	if (addr_offset != 0 && length > 0) {
+		uint8_t temp_buf[4];
+		size_t prefix = f_param->write_block_size - addr_offset;
+
+		if (prefix > length) {
+			prefix = length;
+		}
+
+		memset(temp_buf, f_param->erase_value, sizeof(temp_buf));
+		memcpy(temp_buf + addr_offset, data_ptr, prefix);
+
+		uint32_t aligned_addr = address & ~(f_param->write_block_size - 1);
+
+		dev_data->trans_conf.wait_cycles = 0;
+		dev_data->trans_conf.addr_len = 0;
+
+		ret = alif_hal_ospi_prepare_transfer(dev_data->ospi_handle,
+						     &dev_data->trans_conf);
+		if (ret != 0) {
+			ret = err_map_alif_hal_to_zephyr(ret);
+			goto out;
+		}
+
+		ret = set_write_enable(dev, OSPI_DFS_BITS_16);
+		if (ret != 0) {
+			goto out;
+		}
+
+		dev_data->cmd_buf[0] = CMD_PAGE_PROGRAM;
+		dev_data->cmd_buf[1] = aligned_addr;
+
+		switch (dev_cfg->rw_dfs) {
+		case OSPI_DFS_BITS_8:
+			dev_data->cmd_buf[2] = temp_buf[0];
+			break;
+		case OSPI_DFS_BITS_16: {
+			uint16_t v;
+
+			memcpy(&v, temp_buf, sizeof(v));
+			dev_data->cmd_buf[2] = v;
+			break;
+		}
+		case OSPI_DFS_BITS_32: {
+			uint32_t v;
+
+			memcpy(&v, temp_buf, sizeof(v));
+			dev_data->cmd_buf[2] = v;
+			break;
+		}
+		}
+
+		dev_data->trans_conf.wait_cycles = 0;
+		dev_data->trans_conf.addr_len = OSPI_ADDR_LENGTH_32_BITS;
+		dev_data->trans_conf.frame_size = dev_cfg->rw_dfs;
+
+		ret = alif_hal_ospi_prepare_transfer(dev_data->ospi_handle,
+						     &dev_data->trans_conf);
+		if (ret != 0) {
+			ret = err_map_alif_hal_to_zephyr(ret);
+			goto out;
+		}
+
+		ret = set_cs_pin(dev_data->ospi_handle, SLAVE_ACTIVATE);
+		if (ret != 0) {
+			goto out;
+		}
+
+		k_event_clear(&dev_data->event_f,
+			      OSPI_EVENT_TRANSFER_COMPLETE |
+			      OSPI_EVENT_DATA_LOST);
+
+		ret = alif_hal_ospi_send(dev_data->ospi_handle,
+					 dev_data->cmd_buf, 3);
+		if (ret != 0) {
+			ret = err_map_alif_hal_to_zephyr(ret);
+			set_cs_pin(dev_data->ospi_handle, SLAVE_DE_ACTIVATE);
+			goto out;
+		}
+
+		event = k_event_wait(&dev_data->event_f,
+				     OSPI_EVENT_TRANSFER_COMPLETE |
+				     OSPI_EVENT_DATA_LOST,
+				     false, K_FOREVER);
+		if (!(event & OSPI_EVENT_TRANSFER_COMPLETE)) {
+			LOG_ERR("F-Write: Incomplete Event [%d]", event);
+			set_cs_pin(dev_data->ospi_handle, SLAVE_DE_ACTIVATE);
+			ret = -EIO;
+			goto out;
+		}
+
+		ret = set_cs_pin(dev_data->ospi_handle, SLAVE_DE_ACTIVATE);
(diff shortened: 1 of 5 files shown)


tero-alifsemi (commented):
MCUBoot ospi -> mram case tested and verified.

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -231,11 +245,82 @@ static int flash_is25wx_ospi_read(const struct device *dev, off_t address, void
...
+		ret = alif_hal_ospi_prepare_transfer(dev_data->ospi_handle,
+						     &dev_data->trans_conf);
+		if (ret != 0) {
+			ret = err_map_alif_hal_to_zephyr(ret);
+			k_sem_give(&dev_data->sem);
+			return ret;
+		}
+
+		ret = set_cs_pin(dev_data->ospi_handle, SLAVE_ACTIVATE);
+		if (ret == 0) {
```
tero-alifsemi: Can you changes ret value checks to check for errors, not for OK? Would be more readable and in line with coding style.

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -288,13 +285,58 @@ static int flash_is25wx_ospi_read(const struct device *dev, off_t address, void
...
+				set_cs_pin(dev_data->ospi_handle, SLAVE_DE_ACTIVATE);
+				ret = -EIO;
+			}
+		}
+		if (ret == 0) {
+			ret = set_cs_pin(dev_data->ospi_handle, SLAVE_DE_ACTIVATE);
+		}
+		if (ret == 0) {
+			memcpy(data_ptr, temp_buf, remaining);
+		}
```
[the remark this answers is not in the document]
Tanay-AlifSemi: No overflow risk. write_block_size is constrained to {1, 2, 4} by the get_dfs() function, which maps it to OSPI DFS 8/16/32-bit. This value comes from the device tree and has no other valid values.
Therefore remaining = write_block_size - 1 is at most 3 bytes, which always fits in temp_buf[4].

The transfer intentionally reads/writes exactly 1 block (= 1 OSPI frame = write_block_size bytes). In DDR mode, the OSPI controller cannot transfer fewer than write_block_size bytes. So, reading one full block into temp_buf and copying only the remaining needed bytes is the correct and intended approach.

Unused "read_size" is cleaned up.

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -288,13 +285,58 @@ static int flash_is25wx_ospi_read(const struct device *dev, off_t address, void
...
+		if (ret == 0) {
+			ret = set_cs_pin(dev_data->ospi_handle, SLAVE_ACTIVATE);
+		}
+		if (ret == 0) {
+			cmd[0] = CMD_READ_DATA;
+			cmd[1] = address;
+			k_event_clear(&dev_data->event_f,
+					  OSPI_EVENT_TRANSFER_COMPLETE | OSPI_EVENT_DATA_LOST);
+
+			ret = alif_hal_ospi_transfer(dev_data->ospi_handle, cmd, temp_buf, 1);
```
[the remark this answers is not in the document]
Tanay-AlifSemi: No overflow risk. write_block_size is constrained to {1, 2, 4} by the get_dfs() function, which maps it to OSPI DFS 8/16/32-bit. This value comes from the device tree and has no other valid values.
Therefore remaining = write_block_size - 1 is at most 3 bytes, which always fits in temp_buf[4].

The transfer intentionally reads/writes exactly 1 block (= 1 OSPI frame = write_block_size bytes). In DDR mode, the OSPI controller cannot transfer fewer than write_block_size bytes. So, reading one full block into temp_buf and copying only the remaining needed bytes is the correct and intended approach.

Unused "read_size" is cleaned up.

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -288,13 +285,58 @@ static int flash_is25wx_ospi_read(const struct device *dev, off_t address, void
...
+				ret = -EIO;
+			}
+		}
+		if (ret == 0) {
+			ret = set_cs_pin(dev_data->ospi_handle, SLAVE_DE_ACTIVATE);
+		}
+		if (ret == 0) {
+			memcpy(data_ptr, temp_buf, remaining);
+		}
+	}
```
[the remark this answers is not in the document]
Tanay-AlifSemi: No overflow risk. write_block_size is constrained to {1, 2, 4} by the get_dfs() function, which maps it to OSPI DFS 8/16/32-bit. This value comes from the device tree and has no other valid values.
Therefore remaining = write_block_size - 1 is at most 3 bytes, which always fits in temp_buf[4].

The transfer intentionally reads/writes exactly 1 block (= 1 OSPI frame = write_block_size bytes). In DDR mode, the OSPI controller cannot transfer fewer than write_block_size bytes. So, reading one full block into temp_buf and copying only the remaining needed bytes is the correct and intended approach.

Unused "read_size" is cleaned up.

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -288,13 +285,58 @@ static int flash_is25wx_ospi_read(const struct device *dev, off_t address, void
...
+
+		if (read_size > sizeof(temp_buf)) {
+			read_size = sizeof(temp_buf);
+		}
+
+		dev_data->trans_conf.wait_cycles = 16;
+		dev_data->trans_conf.addr_len = OSPI_ADDR_LENGTH_32_BITS;
+		dev_data->trans_conf.frame_size = dev_cfg->rw_dfs;
+
+		ret = alif_hal_ospi_prepare_transfer(dev_data->ospi_handle, &dev_data->trans_conf);
```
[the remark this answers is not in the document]
Tanay-AlifSemi: No restore needed. The function is protected by a semaphore (k_sem_take), so there is no concurrent access to `dev_data->trans_conf`. Every call to `flash_is25wx_ospi_read() / flash_is25wx_ospi_write() ` sets `trans_conf.wait_cycles`,` trans_conf.addr_len`, and `trans_conf.frame_size` fresh at the top before any transfer. There is no caller that depends on `trans_conf` retaining state across calls. Saving/restoring would be dead code.

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -288,13 +285,58 @@ static int flash_is25wx_ospi_read(const struct device *dev, off_t address, void
...
+		if (ret == 0) {
```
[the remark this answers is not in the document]
Tanay-AlifSemi: `temp_buf[4]` is correctly sized. `write_block_size` is device-tree driven and constrained to {1, 2, 4} by the `get_dfs() `mapping. There is no silent capping — `temp_buf[4]` always holds exactly one full block.
The comment has been improved to explicitly state: "`write_block_size` is 1, 2, or 4, so `temp_buf[4]` is always large enough."

`/* Handle remaining odd bytes if any:
	 * read one full block into a temp buffer,then copy only the needed bytes.
	 * write_block_size is 1, 2, or 4 so temp_buf[4] is always large enough.
	 */`

drivers/flash/flash_ospi_is25wx.c
Tanay-AlifSemi: updated!

tero-alifsemi (approved):

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -320,22 +463,134 @@ static int flash_is25wx_ospi_write(const struct device *dev, off_t address, cons
...
+		ret = set_cs_pin(dev_data->ospi_handle, SLAVE_DE_ACTIVATE);
+		if (ret != 0) {
+			goto out;
+		}
+
+		/* Wait for flash to complete programming */
+		do {
+			status = read_status_reg(dev, CMD_READ_FLAG_STATUS,
+						 &val);
+			if (status != 0) {
```
JuhoEskeli-AlifSemi, line 579: Here the status value is not followed up on (i.e. we break out, but then we don't care)

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -320,22 +463,134 @@ static int flash_is25wx_ospi_write(const struct device *dev, off_t address, cons
...
+		ret = set_write_enable(dev, OSPI_DFS_BITS_16);
+		if (ret != 0) {
+			goto out;
+		}
+
+		dev_data->cmd_buf[0] = CMD_PAGE_PROGRAM;
+		dev_data->cmd_buf[1] = aligned_addr;
+
+		switch (dev_cfg->rw_dfs) {
+		case OSPI_DFS_BITS_8:
```
JuhoEskeli-AlifSemi, line 512: Casting like so is undefined behavior under the strict-aliasing rule. 
E.g.:

switch (dev_cfg->rw_dfs) {
case OSPI_DFS_BITS_8:
    dev_data->cmd_buf[2] = temp_buf[0];
    break;
case OSPI_DFS_BITS_16: {
    uint16_t v;
    memcpy(&v, temp_buf, sizeof(v));
    dev_data->cmd_buf[2] = v;
    break;
}
case OSPI_DFS_BITS_32: {
    uint32_t v;
    memcpy(&v, temp_buf, sizeof(v));
    dev_data->cmd_buf[2] = v;
    break;
}
}

drivers/flash/flash_ospi_is25wx.c
```diff
@@ -448,6 +704,106 @@ static int flash_is25wx_ospi_write(const struct device *dev, off_t address, cons
...
+		if (ret != 0) {
+			goto out;
+		}
+
+		dev_data->cmd_buf[0] = CMD_PAGE_PROGRAM;
+		dev_data->cmd_buf[1] = address;
+
+		/* Pack one word from temp_buf */
+		switch (dev_cfg->rw_dfs) {
+		case OSPI_DFS_BITS_8:
```
JuhoEskeli-AlifSemi, line 748: Note previous comment about pointer cast issue. Apply the same fix

drivers/flash/flash_ospi_is25wx.c
Tanay-AlifSemi: updated!

drivers/flash/flash_ospi_is25wx.c
Tanay-AlifSemi: updated!

drivers/flash/flash_ospi_is25wx.c
Tanay-AlifSemi: updated!

JuhoEskeli-AlifSemi (approved):

PR #476, opened 2026-04-27
drivers: isp: Add API to register callback for is stable status

YogenderKumar-AlifSemi:
Add device specific API to register a callback to the isp driver that is called during the bottom half processing of the ISP. The callback is for the AE isStable status passing to the main application.

This PR depends on: alifsemi/sdk-alif/pull/681
This PR depends on: alifsemi/hal_alif/pull/121

### DIFF
--- a/drivers/video/isp_pico.c
+++ b/drivers/video/isp_pico.c
@@ -884,6 +884,16 @@ static DEVICE_API(video, isp_driver_api) = {
 #endif /* CONFIG_POLL */
 };
 
+inline int z_impl_register_ae_status_callback(const struct device *dev,
+		isp_ae_status_cb ae_status_cb)
+{
+	struct isp_data *data = dev->data;
+
+	data->init_cfg.ae_status_cb = ae_status_cb;
+
+	return 0;
+}
+
 static int isp_configure(const struct device *dev)
 {
 	const struct isp_config *config = dev->config;
--- a/drivers/video/isp_pico.c
+++ b/drivers/video/isp_pico.c
@@ -888,6 +888,28 @@ static DEVICE_API(video, isp_driver_api) = {
 #endif /* CONFIG_POLL */
 };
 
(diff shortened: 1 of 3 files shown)


drivers/video/isp_pico.c
```diff
@@ -884,6 +884,16 @@ static DEVICE_API(video, isp_driver_api) = {
 #endif /* CONFIG_POLL */
 };
 
+inline int z_impl_register_ae_status_callback(const struct device *dev,
```
silesh-alifsemi: Should we add a `z_vrfy_` function too and is the `inline` valid here? https://docs.zephyrproject.org/latest/kernel/drivers/index.html#device-specific-api-extensions.

include/zephyr/drivers/video/isp-vsi.h
```diff
@@ -88,24 +88,40 @@ struct channel_parameters {
...
 };
 
+/**
+ * @brief Callback invoked each frame end with the current AE stability status.
+ *
+ * Called from interrupt bottom-half context (workqueue).
+ *
+ * @param ae_stable  1 if AE has converged, 0 otherwise.
+ */
+typedef void (*isp_ae_status_cb)(uint8_t ae_stable);
```
silesh-alifsemi: The driver callbacks should add `const struct device *dev` as the first argument. Optionally a `void *user_data` pointer too. https://docs.zephyrproject.org/latest/develop/api/design_guidelines.html#using-callbacks

silesh-alifsemi (changes requested):

drivers/video/isp_pico.c
YogenderKumar-AlifSemi: We can add the z_vrfy_ function for the same, but how do we validate the function pointer coming from user-space.  I am not sure, if other than NULL check any other type check can be done on the userspace function pointer.

silesh-alifsemi (approved):

RupeshKumar-AlifSemi (approved):

PR #504, opened 2026-05-18
drivers: dma: pl330: add hardware scatter-gather support

CijaSathish-AlifSemi:
Implement hardware scatter-gather support in the PL330 DMA driver.
When `source_gather_en` is set in the head block, the driver builds a
single PL330 microcode program covering all blocks in the
dma_block_config chain. The hardware executes every block autonomously
and raises one interrupt at the end, with no CPU involvement between
blocks.

Signed-off-by: Cija Sathishkumar <cija.sathishkumar@alifsemi.com>

### DIFF
--- a/drivers/dma/dma_pl330.h
+++ b/drivers/dma/dma_pl330.h
@@ -9,6 +9,7 @@
 
 #include <zephyr/drivers/dma.h>
 #include <zephyr/drivers/clock_control.h>
+#include <zephyr/spinlock.h>
 
 #define DT_DRV_COMPAT arm_dma_pl330
 /*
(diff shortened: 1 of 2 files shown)


drivers/dma/dma_pl330.h
```diff
@@ -9,6 +9,7 @@
 
 #include <zephyr/drivers/dma.h>
 #include <zephyr/drivers/clock_control.h>
+#include <zephyr/spinlock.h>
```
sudhir-alifsemi, line 12: can we remove this header?

sudhir-alifsemi (commented):
In SG transfer, how others are reporting the dma_get_status, do we have reference?

PR #520, opened 2026-05-22
clock_control: alif: balletto: fix HFXO divider bit position

Tanay-AlifSemi:
The CLKDIV field in AON_MISC_REG1 is at bits [20:17] per the HWRM, not bits [16:13] which fall in the reserved region. Reading from the wrong bit position returned a divider value of 7 (from reserved bits POR default), causing get_hfxo_divided_clk_freq() to report 300 kHz instead of the actual 38.4 MHz. This made UART baud rate calculation incorrect — divisor computed as 0 at 115200 baud, hanging the UART.

Fix the shift from >> 13 to >> 17 and add lookup table for divider values 8-15 per HWRM encoding.

### DIFF
--- a/drivers/clock_control/clock_control_alif_balletto.c
+++ b/drivers/clock_control/clock_control_alif_balletto.c
@@ -77,6 +77,10 @@ struct clock_control_alif_config {
 /** clock module (from clkid cell) */
 #define ALIF_CLOCK_CFG_MODULE(id) (((id) >> ALIF_CLOCK_MODULE_SHIFT) & ALIF_CLOCK_MODULE_MASK)
 
+/* AON MISC_REG1: HFXO clock divider bit position */
+#define AON_MISC_REG1_HFXO_CLKDIV_POS	17U
+#define AON_MISC_REG1_HFXO_CLKDIV_MASK	0xFU
+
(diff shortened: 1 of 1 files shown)


arashed-alif (approved):
fine

petrih-alifsemi (approved):

drivers/clock_control/clock_control_alif_balletto.c
```diff
@@ -159,10 +159,33 @@ static uint32_t get_syst_pclk_freq(void)
 
 static uint32_t get_hfxo_divided_clk_freq(void)
 {
-	uint32_t divider = (sys_read32(AON_MISC_REG1) >> 13) & 0xF;
+	const uint32_t divider = (sys_read32(AON_MISC_REG1) >> 17) & 0xF;
```
sudhir-alifsemi: Instead of magic number 17, would be better to have a readable macro.

petrih-alifsemi (approved):

sudhir-alifsemi (approved):

drivers/clock_control/clock_control_alif_balletto.c
Tanay-AlifSemi: updated.

PR #522, opened 2026-05-25
boards: alif: common: move sd regulator to root node

Deepak-AlifSemi:
Move the SD-regulator node entry from the soc node to the root node to avoid a warning.


RupeshKumar-AlifSemi (approved):

sudhir-alifsemi (commented):
It would be better to mention the warning in the description

Deepak-AlifSemi:
> It would be better to mention the warning in the description

done

sudhir-alifsemi (approved):

PR #532, opened 2026-05-26
drivers: clock_control: force PCLK on for UART DMA on E8/E1C/B1

Tanay-AlifSemi:
UART DMA transfers on Ensemble Gen2 (E8), E1C, and Balletto (B1) require the APB interface clock (PCLK) to be force-enabled via the EXPMST0 control register. Without this, DMA TX may stall.

Gate the PCLK force behind CONFIG_UART_ASYNC_API so it is only applied when DMA mode is in use. IPCLK force is not needed.

On E7, PCLK force is applied unconditionally for all UART modes.

### DIFF
--- a/drivers/clock_control/clock_control_alif_balletto.c
+++ b/drivers/clock_control/clock_control_alif_balletto.c
@@ -81,6 +81,10 @@ struct clock_control_alif_config {
 #define AON_MISC_REG1_HFXO_CLKDIV_POS	17U
 #define AON_MISC_REG1_HFXO_CLKDIV_MASK	0xFU
 
+/* EXPMST0 control register config */
+#define ALIF_EXPMST0_CTRL_IPCLK_FORCE_BIT BIT(31U)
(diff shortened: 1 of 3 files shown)


kjbracey (approved):
This activates whenever the DMA-using APIs exist, rather than when they're actually used, which would be even better, but this is certainly an improvement over always enabling the force - gives you a build option that avoids it.

sudhir-alifsemi (approved):

kjbracey:
Belatedly I've remembered we had a better workaround for this, which was to disable the low-power signalling in the UART itself - clear LCR_EXT.CGEN. That avoids forcing clocks for other peripherals, and would be something that could be done in the UART driver only when the DMA is used.

Tanay-AlifSemi:
> Belatedly I've remembered we had a better workaround for this, which was to disable the low-power signalling in the UART itself - clear LCR_EXT.CGEN. That avoids forcing clocks for other peripherals, and would be something that could be done in the UART driver only when the DMA is used.

sure, I will check this.

PR #542, opened 2026-05-28
video: arx3a0: Set default format in arx3a0 init

Shivakumar-AlifSemi:
video_get_fmt fails when called before video_set_fmt. Set default format during arx3a0's init to support video_get_fmt call

### DIFF
--- a/drivers/video/arx3a0.c
+++ b/drivers/video/arx3a0.c
@@ -48,6 +48,11 @@ LOG_MODULE_REGISTER(arx3a0);
 
 #define ARX3A0_FRAME_RATE 5
 
+/* Default format settings for ARX3A0 (RAW10 MIPI 560x560) */
+#define ARX3A0_DEFAULT_WIDTH   560
+#define ARX3A0_DEFAULT_HEIGHT  560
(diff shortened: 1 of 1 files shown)


YogenderKumar-AlifSemi (approved):
Looks Good!

drivers/video/arx3a0.c
```diff
@@ -1337,6 +1337,23 @@ static int arx3a0_init(const struct device *dev)
...
 
+	/*
+	 * Currently ARX3A0 sensor supports only one format - RAW10 MIPI.
+	 * Set default format so video_get_format works before video_set_format.
+	 */
+	struct video_format fmt = {
+		.pixelformat = VIDEO_PIX_FMT_Y10P,
+		.width = 560,
+		.height = 560,
+		.pitch = 560 << 1,
```
RupeshKumar-AlifSemi: can use macros instead of hard coded values.

RupeshKumar-AlifSemi (changes requested):

drivers/video/arx3a0.c
Shivakumar-AlifSemi: Comment is addressed!

RupeshKumar-AlifSemi (approved):

PR #545, opened 2026-06-02
drivers: audio: alif_pdm: add proper PM suspend/resume support

AravindKrishnan-AlifSemi:
Add dedicated pdm_suspend() and pdm_resume() functions instead of re-running the full init on resume. This ensures correct power management behavior:

- pdm_suspend(): disables interrupts, turns off the PDM clock, and applies the PINCTRL_STATE_SLEEP pin configuration.
- pdm_resume(): restores pin state to default, reconfigures and enables the clock, and re-initializes PDM filter/threshold registers.

Additionally:
- Mark the device busy on DMIC_TRIGGER_START and clear on DMIC_TRIGGER_STOP to prevent the PM framework from suspending the device during active audio capture.
- Return -ENOTSUP for unsupported PM actions instead of falling through to it.

### DIFF
--- a/drivers/audio/alif_pdm.c
+++ b/drivers/audio/alif_pdm.c
@@ -274,6 +274,8 @@ static int dmic_alif_pdm_trigger(const struct device *dev, enum dmic_trigger cmd
 		while (k_msgq_get(&pdata->buf_queue, &buf, K_NO_WAIT) == 0) {
 			k_mem_slab_free(pdata->mem_slab, buf);
 		}
+
+		pm_device_busy_clear(dev);
 		break;
 
 	case DMIC_TRIGGER_START:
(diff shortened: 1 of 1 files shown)


drivers/audio/alif_pdm.c
```diff
@@ -577,7 +581,75 @@ static const struct _dmic_ops dmic_alif_pdm_api = {
...
+			return ret;
+		}
+	}
 
+#if defined(CONFIG_PINCTRL)
+	/* Apply sleep pin configuration if available */
+	if (cfg->pcfg != NULL) {
+		ret = pinctrl_apply_state(cfg->pcfg, PINCTRL_STATE_SLEEP);
+		if (ret < 0 && ret != -ENOENT) {
+			/* Ignore -ENOENT (sleep state not defined) */
```
silesh-alifsemi, line 604: This comment should go above the `if (ret < 0 && ret != -ENOENT) {`

silesh-alifsemi (approved):

RupeshKumar-AlifSemi (approved):

PR #549, opened 2026-06-03
fix clock divisor rounding in set_rate

JuhoEskeli-AlifSemi:
This pull request updates how clock frequency dividers are calculated in three different clock control drivers. Instead of using simple integer division, the code now uses the `DIV_ROUND_CLOSEST` macro to ensure the computed divider is rounded to the nearest integer, which should improve the accuracy of frequency settings.

**Clock divider calculation improvements:**

* Updated the divider calculation in `alif_clock_control_set_rate` to use `DIV_ROUND_CLOSEST` instead of plain integer division in the following files:
  * `clock_control_alif_balletto.c`
  * `clock_control_alif_ensemble.c`
  * `clock_control_alif_ensemble_e1c.c`

### DIFF
--- a/drivers/clock_control/clock_control_alif_balletto.c
+++ b/drivers/clock_control/clock_control_alif_balletto.c
@@ -559,7 +559,7 @@ static int alif_clock_control_set_rate(const struct device *dev, clock_control_s
 	}
 	reg_addr = module_base + ALIF_CLOCK_CFG_REG(clk_id);
 
-	freq_div = (clk_freq / frequency);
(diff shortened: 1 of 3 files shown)


KimmoVaisanen-AlifSemi (approved):

sudhir-alifsemi (commented):
Please update it for balletto too.

JuhoEskeli-AlifSemi:
Updated @sudhir-alifsemi

sudhir-alifsemi (approved):

jheiskan81 (approved):

petrih-alifsemi (approved):

PR #552, opened 2026-06-08
drivers: adc: alif: Correct buffer indexing and add scan mode validation

Prabhakar-Alifsemi:
Update check_buffer_size to size the buffer based on the highest channel ID for continuous mode, and on a single sample for single-shot mode.

Add a validation guard in adc_set_ch_scan_mode to reject the unsupported combination of single-shot conversion with multi-channel scan.

### DIFF
--- a/drivers/adc/adc_alif.c
+++ b/drivers/adc/adc_alif.c
@@ -437,15 +437,24 @@ static inline void adc_init_channel_select(uintptr_t adc, uint32_t channel)
 	sys_write32(data, adc + ADC_SEQUENCER_CTRL);
 }
 
-static inline void adc_set_ch_scan_mode(const struct device *dev)
+static inline int adc_set_ch_scan_mode(const struct device *dev)
 {
 	uint32_t regs = DEVICE_MMIO_NAMED_GET(dev, adc_reg);
 	const struct adc_config *config = dev->config;
 	uint32_t data;
 
+	/* Error: single-shot conversion with multi-channel scan is not supported */
+	if ((config->conv_mode == ADC_CONV_MODE_SINGLE_SHOT) &&
+	    (config->scan_mode == ADC_SCAN_MODE_MULTI_CH)) {
+		LOG_ERR("Single-shot conversion with multi-channel scan mode is not supported");
+		return -EINVAL;
+	}
+
 	data = sys_read32(regs + ADC_SEQUENCER_CTRL);
 	data |= config->scan_mode;
 	sys_write32(data, regs + ADC_SEQUENCER_CTRL);
+
+	return 0;
 }
 
 static inline void enable_adc24(const struct device *dev)
@@ -653,11 +662,18 @@ void adc_context_update_buffer_pointer(struct adc_context *ctx, bool repeat_samp
 	}
 }
 
-static int check_buffer_size(const struct adc_sequence *sequence, uint8_t active_channels)
+static int check_buffer_size(const struct adc_sequence *sequence,
+			     uint8_t active_channels, uint8_t conv_mode)
 {
 	size_t needed_buffer_size;
 
-	needed_buffer_size = active_channels * sizeof(uint32_t);
+	if (conv_mode == ADC_CONV_MODE_SINGLE_SHOT) {
+		/* Single-shot: single sample per conversion */
+		needed_buffer_size = sizeof(uint32_t);
+	} else {
+		/* Continuous/multi-channel: buffer size based on number of active channels */
+		needed_buffer_size = (size_t)active_channels * sizeof(uint32_t);
+	}
 
 	if (sequence->options) {
 		needed_buffer_size *= (1 + sequence->options->extra_samplings);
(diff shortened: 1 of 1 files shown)


KimmoVaisanen-AlifSemi (approved):

drivers/adc/adc_alif.c
```diff
@@ -995,7 +1020,10 @@ static int adc_init(const struct device *dev)
 	adc_set_comparator_ctrl_bit(regs, config->comparator_threshold_comparison);
 
 	/* set channel scan mode */
-	adc_set_ch_scan_mode(dev);
+	ret = adc_set_ch_scan_mode(dev);
```
silesh-alifsemi: `ret` not declared ?

silesh-alifsemi (changes requested):

drivers/adc/adc_alif.c
```diff
@@ -523,8 +532,9 @@ void adc_done0_irq_handler(const struct device *dev)
 		/* storing the address to be fetched for particular channels */
 		channel_sample_reg = sample_reg + sizeof(uint32_t) * channel;
 
-		/* storing the digital output to the respective buffer cells */
-		*(data->buffer + data->curr_cnt++) = sys_read32(channel_sample_reg);
+		/* store digital output in buffer indexed by channel id */
+		*(data->buffer + channel) = sys_read32(channel_sample_reg);
```
AnttiKauppila-AlifSemi: How well have you tested this? You seem to change the behaviour from "ring buffer style" approach to channel based approach. I am not familiar with this but is it possible that there are situation when channel x data is waiting to be handled and new data is received for channel x (overriding the previous data)?
Also why is adc_done1_irq_handler() not updated similarly with this? If that gets updated -> curr_cnt loses its meaning and can be thus removed.

drivers/adc/adc_alif.c
Prabhakar-Alifsemi: Done

drivers/adc/adc_alif.c
Prabhakar-Alifsemi: adc_done1_irq_handler() only handles single-shot, single-channel conversions, as this PR rejects single-shot + multi-channel scan. In this case, curr_cnt is always 0, making buffer[0] and buffer[channel] effectively equivalent. Hence, no update to this handler is necessary.

Yes, I agree. The ring buffer style is the correct approach since it preserves the expected Zephyr ADC semantics, where samples are stored sequentially in the order of the configured channels rather than indexed by channel number.

drivers/adc/adc_alif.c
```diff
@@ -644,11 +653,25 @@ void adc_context_update_buffer_pointer(struct adc_context *ctx, bool repeat_samp
...
 
-	needed_buffer_size = active_channels * sizeof(uint32_t);
+	if (conv_mode == ADC_CONV_MODE_SINGLE_SHOT) {
+		/* Single-shot: buffer size based on active channels count */
+		needed_buffer_size = sizeof(uint32_t);
+	} else {
+		/* Continuous/multi-channel: buffer size based on highest channel id */
+		uint8_t highest_channel = 0U;
+
+		for (uint8_t ch = 0; ch < ADC_LAST_AVAILABLE_CHANNEL + 1; ch++) {
```
sudhir-alifsemi: count the number of active channels rather than the highest channel.

sudhir-alifsemi (changes requested):
Please update the git commit msg accordingly

drivers/adc/adc_alif.c
Prabhakar-Alifsemi: updated the buffer allocation based on the number of channels.

sudhir-alifsemi (approved):

BhupeshKharwa-AlifSemi:
Can someone please approve this so that the minimum requirement is satisfied and this gets merged. This is a bit urgent as customer is waiting for merge of the patch that they provided it is has been about a month. Thank you.

drivers/adc/adc_alif.c
```diff
@@ -987,7 +1004,10 @@ static int adc_hw_init(const struct device *dev)
 	adc_set_comparator_ctrl_bit(regs, config->comparator_threshold_comparison);
 
 	/* set channel scan mode */
-	adc_set_ch_scan_mode(dev);
+	err = adc_set_ch_scan_mode(dev);
```
KimmoVaisanen-AlifSemi, line 1007: err is undefined

KimmoVaisanen-AlifSemi (changes requested):

drivers/adc/adc_alif.c
Prabhakar-Alifsemi: addressed

KimmoVaisanen-AlifSemi: Was it? I still don't see declaration of `err`.

Prabhakar-Alifsemi: err is defined now

sudhir-alifsemi (approved):

KimmoVaisanen-AlifSemi (approved):

silesh-alifsemi (approved):

PR #555, opened 2026-06-11
drivers: misc: ethos_u: add PM support

jarvte:
Add PM support for ethos driver

### DIFF
--- a/drivers/misc/ethos_u/ethos_u.c
+++ b/drivers/misc/ethos_u/ethos_u.c
@@ -11,8 +11,10 @@
 #include <zephyr/kernel.h>
 #include <zephyr/irq.h>
 #include <zephyr/sys/util.h>
-
+#include <zephyr/pm/device.h>
+#include <zephyr/pm/policy.h>
 #include <ethosu_driver.h>
+#include <ethosu_device.h>
 
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(ethos_u, CONFIG_ARM_ETHOS_U_LOG_LEVEL);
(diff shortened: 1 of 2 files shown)


drivers/misc/ethos_u/ethos_u.c
```diff
@@ -150,6 +152,45 @@ static int ethosu_zephyr_init(const struct device *dev)
...
+ */
+static int ethosu_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	int ret = 0;
+	struct ethosu_data *data = dev->data;
+	struct ethosu_driver *drv = &data->drv;
+
+	switch (action) {
+	case PM_DEVICE_ACTION_RESUME:
+		return ethosu_zephyr_init(dev);
```
AravindKrishnan-AlifSemi, line 187: This needs to be checked, a complete init is not required, whatever the software context like mutex or semaphore states are residing in the retention enabled memory, only the hardware changes, power and clock management is needed here.

AravindKrishnan-AlifSemi:
pm_device_busy_set and clear is not handled

AravindKrishnan-AlifSemi:
https://github.com/alifsemi/zephyr_alif/pull/578
@jarvte in the above-mentioned PR I have added one clock support for U85 on our ensemble gen2. You can consider this also in yours suspend and resume part.

drivers/misc/ethos_u/ethos_u.c
jarvte: mutex and semaphore already has a guard so they are not initialized again: https://gitlab.arm.com/artificial-intelligence/ethos-u/ethos-u-core-driver/-/blob/main/src/ethosu_driver.c?ref_type=heads#L425
https://gitlab.arm.com/artificial-intelligence/ethos-u/ethos-u-core-driver/-/blob/main/src/ethosu_driver.c?ref_type=heads#L435

ethosu_driver.c and files in that folder are not included in Zephyr but coming from https://gitlab.arm.com/artificial-intelligence/ethos-u/ethos-u-core-driver so it's can't be changed not that easily.

jarvte:
> pm_device_busy_set and clear is not handled

@AravindKrishnan-AlifSemi  I don't think they are needed as device is going off. Only should be used when doing some transaction like writing to flash etc to prevent sleep. Correct me if I'm wrong as I'm not that familiar with Zephyr yet?

tero-alifsemi (approved):

JuhoEskeli-AlifSemi (approved):

PR #558, opened 2026-06-11
drivers: flash: add PM support for is25wx

jarvte:
Add PM support for is25wx

### DIFF
--- a/drivers/flash/flash_ospi_is25wx.c
+++ b/drivers/flash/flash_ospi_is25wx.c
@@ -13,6 +13,7 @@
 #include <zephyr/drivers/pinctrl.h>
 #include <zephyr/irq.h>
 #include <zephyr/kernel.h>
+#include <zephyr/pm/device.h>
 
 #include "flash_ospi_is25wx.h"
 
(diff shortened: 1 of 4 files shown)


GaneshRamani-AlifSemi:
@jarvte , 
 General comment : 
   1. do not we need to handle the device in deep power state ?
   
image

   2. we need to address ospi controller as well get into PM state, but currently the OSPI driver inside HAL which need to migrate into zephyr so that we can integrate system managed PM.

jarvte:
> @jarvte , General comment :
> 
> 1. do not we need to handle the device in deep power state ?
> 
> 2. we need to address ospi controller as well get into PM state, but currently the OSPI driver inside HAL which need to migrate into zephyr so that we can integrate system managed PM.

1) & 2) I don't know, could/should this be done in phases? This was needed to get device to STOP mode and wake up properly to access ML model in OSPI Flash, meaning support for suspend to RAM. @sudhir-alifsemi what is your take on this?

KariHaapalehto-AlifSemi:
@GaneshRamani-AlifSemi @sudhir-alifsemi @petrih-alifsemi 
Please review

petrih-alifsemi (approved):

JuhoEskeli-AlifSemi (approved):

PR #562, opened 2026-06-15
soc: alif: ensemble: add heap move to sram support

jarvte:
Added support to move system heap to SRAM

### DIFF
--- a/soc/alif/ensemble/Kconfig.soc
+++ b/soc/alif/ensemble/Kconfig.soc
@@ -11,4 +11,26 @@ config NONSECURE0_REGION
 	help
 	    Provide config Non-Secure region
 
+choice SYSTEM_HEAP_LOCATION
+	prompt "System heap location"
+	default SYSTEM_HEAP_ZEPHYR_DEFAULT if SOC_FAMILY_ENSEMBLE
+	help
+		Select the SRAM region to place the system heap for Ensemble SoCs.
+		If no option is selected, the system heap will be placed in the default
+		location.
+
+config SYSTEM_HEAP_ZEPHYR_DEFAULT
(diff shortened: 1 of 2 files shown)


JuhoEskeli-AlifSemi (approved):

KariHaapalehto-AlifSemi:
@silesh-alifsemi please review

soc/alif/ensemble/Kconfig.soc
```diff
@@ -11,4 +11,12 @@ config NONSECURE0_REGION
 	help
 	    Provide config Non-Secure region
 
+config SYSTEM_HEAP_IN_SRAM
+	int "Move system heap to SRAM"
+	range -1 1
```
silesh-alifsemi: can we use a `choice` instead of an integer for this target selector? Something like:

https://github.com/alifsemi/zephyr_alif/blob/fe6ffa40653e74c4304ea4106c39f9e9fb7e1e02/soc/nxp/imx/imx8m/Kconfig#L99-L108

That will simplify the `#ifdef` lines in the linker script too. From the same example:

https://github.com/alifsemi/zephyr_alif/blob/fe6ffa40653e74c4304ea4106c39f9e9fb7e1e02/soc/nxp/imx/imx8m/m7/linker.ld#L12-L16

silesh-alifsemi (changes requested):

soc/alif/ensemble/Kconfig.soc
jarvte: changed to use choice

silesh-alifsemi (approved):

sudhir-alifsemi (approved):

petrih-alifsemi (approved):

PR #565, opened 2026-06-16
drivers: can: fix tx callback handling

DelnaDavis-Alifsemi:
…sions

Fixed callback handling to ensure proper calling for each queued frame.


Shreehari-AlifSemi (approved):
Looks okay

silesh-alifsemi (approved):

PR #576, opened 2026-06-30
drivers: display: cdc200: add vsync wait API and fix init ordering

KariHaapalehto-AlifSemi:
Add cdc200_wait_vsync() that blocks on a semaphore signalled by the CDC200 LINE IRQ when a pending framebuffer swap is processed at vblank. This allows DIRECT-mode flush callbacks to synchronise page-flips with the display refresh, preventing tearing and ensuring the back buffer is released before LVGL renders into it again.

Fix the LVGL module init sequence: call lv_init() before registering the log print callback, otherwise lv_init() resets the global state and silently clears the callback pointer.

Fix compile warning about uninitialised variable and unchecked return value in the Alif Ensemble clock control driver (alif_get_module_base).

### DIFF
--- a/drivers/clock_control/clock_control_alif_ensemble.c
+++ b/drivers/clock_control/clock_control_alif_ensemble.c
@@ -326,7 +326,7 @@ static int alif_clock_control_on(const struct device *dev,
 			clock_control_subsys_t sub_system)
 {
 	uint32_t clk_id = (uint32_t) sub_system;
-	uint32_t module_base, reg_addr;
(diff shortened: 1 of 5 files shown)


KariHaapalehto-AlifSemi:
@Shivakumar-AlifSemi @ChandraBhushan-AlifSemi @YogenderKumar-AlifSemi 
My Oulu colleague's are OoO, so could you please review this

petrih-alifsemi (approved):

YogenderKumar-AlifSemi (approved):

PR #581, opened 2026-07-01
drivers: crc: alif_crc: add PM suspend/resume support

Madhumathi-Alifsemi:
Add CRC PM support with dedicated suspend/resume callbacks.

- crc_suspend(): clears CRC_CONTROL[INIT] before suspend.
- crc_resume(): restores seed/custom polynomial configuration and reinitializes the CRC engine.
- Add pm_device_busy_set()/pm_device_busy_clear() around CRC computation to prevent suspend during active operations.
- Return -ENOTSUP for unsupported PM actions.

### DIFF
--- a/drivers/crc/alif_crc.c
+++ b/drivers/crc/alif_crc.c
@@ -357,6 +357,8 @@ static int alif_crc_compute(const struct device *dev, struct crc_params *params)
 		return -EINVAL;
 	}
 
+	pm_device_busy_set(dev);
+
 	/* Initializing crc parameters  */
 	crc_params_init(dev, params);
 
@@ -424,8 +426,14 @@ static int alif_crc_compute(const struct device *dev, struct crc_params *params)
 
 	default:
 		LOG_ERR("Unsupported CRC algorithm");
+
+		pm_device_busy_clear(dev);
+
 		return -EINVAL;
 	}
+
+	pm_device_busy_clear(dev);
+
 	return 0;
 }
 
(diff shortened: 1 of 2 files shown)


drivers/crc/alif_crc.c
```diff
@@ -426,6 +430,11 @@ static int alif_crc_compute(const struct device *dev, struct crc_params *params)
 		LOG_ERR("Unsupported CRC algorithm");
 		return -EINVAL;
 	}
```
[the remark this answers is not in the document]
Madhumathi-Alifsemi: Addressed

drivers/crc/alif_crc.c
```diff
@@ -357,6 +357,10 @@ static int alif_crc_compute(const struct device *dev, struct crc_params *params)
...
 
+	#ifdef CONFIG_PM_DEVICE
```
silesh-alifsemi: You do not need to wrap pm_device_busy_set and pm_device_busy_clear with `#ifdef CONFIG_PM_DEVICE`. They automatically become no-ops with `CONFIG_PM_DEVICE` disabled.

https://github.com/alifsemi/zephyr_alif/blob/79a6c509eafa0a438dfa36cc18fb8a411d03422e/include/zephyr/pm/device.h#L657-L664

silesh-alifsemi (changes requested):

drivers/crc/alif_crc.c
Madhumathi-Alifsemi: Addressed

silesh-alifsemi (approved):

drivers/crc/alif_crc.c
```diff
@@ -424,8 +426,14 @@ static int alif_crc_compute(const struct device *dev, struct crc_params *params)
 
 	default:
 		LOG_ERR("Unsupported CRC algorithm");
+
+		pm_device_busy_clear(dev);
+
 		return -EINVAL;
 	}
+
+		pm_device_busy_clear(dev);
```
sudhir-alifsemi: fix the alignment

drivers/crc/alif_crc.c
```diff
@@ -485,6 +501,63 @@ static int crc_initialize(const struct device *dev)
...
+	return 0;
+}
+
+static int crc_resume(const struct device *dev)
+{
+	struct crc_data *data = dev->data;
+	uintptr_t reg_base = DEVICE_MMIO_GET(dev);
+	uint32_t val;
+
+	if (data->seed_set) {
```
sudhir-alifsemi: do we need explicit seed_set and custom_poly_set flags. Cant we just cache these 2 regs while going to suspend state and restore in the resume path?

sudhir-alifsemi (changes requested):

drivers/crc/alif_crc.c
Madhumathi-Alifsemi: Addressed

RupeshKumar-AlifSemi (approved):

sudhir-alifsemi (approved):

PR #582, opened 2026-07-01
Utimer counter pm support

Kavya-AlifSemi:
Added PM support for UTIMER Counter driver.

### DIFF
--- a/drivers/counter/counter_alif_utimer.c
+++ b/drivers/counter/counter_alif_utimer.c
@@ -9,6 +9,7 @@
 #include <zephyr/irq.h>
 #include <zephyr/drivers/clock_control.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/pm/device.h>
 
 #include "utimer.h"
 #include <zephyr/dt-bindings/timer/alif_utimer.h>
@@ -31,6 +32,10 @@ struct counter_alif_utimer_data {
 	void *top_user_data;
 	atomic_t cc_int_pending;
 	struct counter_alif_utimer_ch_data alarm[NUM_CHANNELS];
+	/* Tracks whether the counter was running before a PM suspend, so that
+	 * pm_resume can correctly restart it only if it was actually running.
+	 */
+	bool running;
 };
 
 struct counter_alif_utimer_config {
@@ -69,20 +74,24 @@ static int utimer_set_direction(uint32_t reg_base, uint8_t direction)
 static int counter_alif_utimer_start(const struct device *dev)
 {
 	const struct counter_alif_utimer_config *cfg = DEV_CFG(dev);
+	struct counter_alif_utimer_data *data = DEV_DATA(dev);
 	uintptr_t global_base = DEVICE_MMIO_NAMED_GET(dev, global);
 
 	/* start the timer counter*/
 	alif_utimer_start_counter(global_base, cfg->timer_id);
+	data->running = true;
 	return 0;
 }
 
 static int counter_alif_utimer_stop(const struct device *dev)
 {
 	const struct counter_alif_utimer_config *cfg = DEV_CFG(dev);
+	struct counter_alif_utimer_data *data = DEV_DATA(dev);
 	uintptr_t global_base = DEVICE_MMIO_NAMED_GET(dev, global);
 
 	/* stop the timer counter */
 	alif_utimer_stop_counter(global_base, cfg->timer_id);
+	data->running = false;
 
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


drivers/counter/counter_alif_utimer.c
```diff
@@ -434,9 +443,72 @@ static int counter_alif_utimer_init(const struct device *dev)
...
+	if (ret != 0) {
+		LOG_ERR("Failed to restore counter direction: err:%d", ret);
+		return ret;
+	}
+
+	alif_utimer_enable_soft_counter_ctrl(timer_base);
+
+	if (data->top_cb) {
+		alif_utimer_enable_interrupt(timer_base, CHAN_INTERRUPT_OVER_FLOW_BIT);
+	}
```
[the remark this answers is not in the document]
Kavya-AlifSemi: updated

drivers/counter/counter_alif_utimer.c
```diff
@@ -31,6 +33,7 @@ struct counter_alif_utimer_data {
 	void *top_user_data;
 	atomic_t cc_int_pending;
 	struct counter_alif_utimer_ch_data alarm[NUM_CHANNELS];
+	bool running;
```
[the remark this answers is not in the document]
Kavya-AlifSemi: updated

drivers/counter/counter_alif_utimer.c
```diff
@@ -10,6 +10,8 @@
 #include <zephyr/drivers/clock_control.h>
 #include <zephyr/logging/log.h>
 
+#include <zephyr/pm/device.h>
```
silesh-alifsemi, line 12: You do not need the blank line above. You can group it with other zephyr includes above.

drivers/counter/counter_alif_utimer.c
```diff
@@ -31,6 +33,10 @@ struct counter_alif_utimer_data {
 	void *top_user_data;
 	atomic_t cc_int_pending;
 	struct counter_alif_utimer_ch_data alarm[NUM_CHANNELS];
+	/* Tracks whether the counter was running before a PM suspend, so that
+	 * pm_resume can correctly restart it only if it was actually running.
+	 */
+	atomic_t running;
```
silesh-alifsemi: This is not manipulated with atomic API below (is treated as a plain `bool`, for eg. `data->running = true;`). I think this can just be a `bool`.

silesh-alifsemi (changes requested):

drivers/counter/counter_alif_utimer.c
Kavya-AlifSemi: updated.

silesh-alifsemi (approved):

drivers/counter/counter_alif_utimer.c
```diff
@@ -434,9 +443,73 @@ static int counter_alif_utimer_init(const struct device *dev)
...
+	}
+
+	alif_utimer_enable_soft_counter_ctrl(timer_base);
+
+	if (data->top_cb) {
+		alif_utimer_clear_interrupt(timer_base, CHAN_INTERRUPT_OVER_FLOW_BIT);
+		alif_utimer_enable_interrupt(timer_base, CHAN_INTERRUPT_OVER_FLOW_BIT);
+	}
+
+	cfg->irq_config(dev);
```
sudhir-alifsemi, line 485: No need to enable the NVIC, that will be restored by the PM framework.

drivers/counter/counter_alif_utimer.c
```diff
@@ -434,9 +443,73 @@ static int counter_alif_utimer_init(const struct device *dev)
...
+
+	alif_utimer_enable_timer_clock(global_base, cfg->timer_id);
+
+	ret = utimer_set_direction(timer_base, cfg->counterdirection);
+	if (ret != 0) {
+		LOG_ERR("Failed to restore counter direction: err:%d", ret);
+		return ret;
+	}
+
+	alif_utimer_enable_soft_counter_ctrl(timer_base);
```
sudhir-alifsemi, line 478: we should enable the clock which is done in the init function,
clock_control_on and similarly we should disable it while going to suspend state

sudhir-alifsemi (changes requested):

drivers/counter/counter_alif_utimer.c
Kavya-AlifSemi: removed.

drivers/counter/counter_alif_utimer.c
Kavya-AlifSemi: updated.

sudhir-alifsemi (approved):

PR #583, opened 2026-07-01
Usb device ncm support

Deepak-AlifSemi:
Added USB device_next NCM class support for alif boards.


subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -229,6 +229,9 @@ struct cdc_ncm_eth_data {
 	enum iface_state if_state;
 	uint16_t tx_seq;
 	uint16_t rx_seq;
+	uint16_t packet_filter;
+	uint16_t ntb_input_size;
```
sudhir-alifsemi: Should this be uint32_t ?

subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -888,20 +891,76 @@ static int usbd_cdc_ncm_ctd(struct usbd_class_data *const c_data,
...
-			LOG_DBG("bRequest 0x%02x (%s) not implemented",
-				setup->bRequest, "SetNtbInputSize");
+			const struct device *dev = usbd_class_get_private(c_data);
+			struct cdc_ncm_eth_data *data = dev->data;
+			const struct ntb_input_size *ntb;
+
+			if (buf == NULL || buf->len < sizeof(struct ntb_input_size)) {
+				LOG_ERR("SetNtbInputSize: invalid data payload");
+				errno = -ENOTSUP;
+				return 0;
```
sudhir-alifsemi, line 934: Should we return success even after setting the errno?

subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -888,20 +891,76 @@ static int usbd_cdc_ncm_ctd(struct usbd_class_data *const c_data,
...
-			LOG_DBG("bRequest 0x%02x (%s) not implemented",
-				setup->bRequest, "SetPacketFilter");
+			const struct device *dev = usbd_class_get_private(c_data);
+			struct cdc_ncm_eth_data *data = dev->data;
+
+			LOG_DBG("bRequest 0x%02x (SetPacketFilter) wValue 0x%04x",
+				setup->bRequest, setup->wValue);
+
+			/* Store packet filter value */
+			data->packet_filter = setup->wValue;
```
sudhir-alifsemi, line 901: Is this variable consumed elsewhere?

subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -888,20 +891,76 @@ static int usbd_cdc_ncm_ctd(struct usbd_class_data *const c_data,
...
+
+			/* Enable interface when packet filter is set */
+			if (setup->wValue != 0) {
+				atomic_set_bit(&data->state, CDC_NCM_IFACE_UP);
+				net_if_carrier_on(data->iface);
+				LOG_INF("CDC NCM interface enabled");
+			} else {
+				atomic_clear_bit(&data->state, CDC_NCM_IFACE_UP);
+				net_if_carrier_off(data->iface);
```
sudhir-alifsemi, line 910: Should we disconnect the link for the '0' case?

sudhir-alifsemi (changes requested):

subsys/usb/device_next/class/usbd_cdc_ncm.c
Deepak-AlifSemi: Yes, this is the correct convention because:
It returns Success as the request is handled but not supported, instead of throwing a transport-level error. The caller checks errno, not the return value, and it is the same throughout the file.

subsys/usb/device_next/class/usbd_cdc_ncm.c
Deepak-AlifSemi: ntb_input_size updated to uint32_t, packet_filter no need to change.

subsys/usb/device_next/class/usbd_cdc_ncm.c
Deepak-AlifSemi: It preserves the host-configured filter value for future use (e.g., for implementing GET_ETHERNET_PM_FILTER or actual packet filtering) and aligns with the CDC NCM spec, which expects the device to track this setting.

subsys/usb/device_next/class/usbd_cdc_ncm.c
Deepak-AlifSemi: The 0 case correctly does this: clearing CDC_NCM_IFACE_UP and calling net_if_carrier_off brings the link down.

RupeshKumar-AlifSemi (approved):
Looks OK

sudhir-alifsemi (approved):

PR #593, opened 2026-07-17
upstream driver: ospi and is25wx driver

GaneshRamani-AlifSemi:
Summary :

1. This commit has backported activity for octal spi controller and IS25Wx device.
2. Both driver taken from main to avoid duplicate and easy for upstream migration.
3. In current ZAS state, the ospi driver in HAL layer which can't support Power-management implementation.
4. IS25Wx driver also ported with DDR support to replace Alif implemented IS25wx driver.
5. Added necessary cleaning for driver property compatibility and created Alif binding to include Alif SoC Specific feature (AES, XiP)
6. Added upstream IS25ws Flash driver support over all the boards. (except E8-ak, other engineering specific test boards)
7. Remaining activity :
    1. PM.
    2. DMA.
    3. Delete Alif IS25wx driver (once the review completed)

Other pending activity to remove HAL OSPI driver from ZAS:
    1. Add Macronix support (upstream available)
    2. RAM support (HyperRAM, PS RAM (APS Memmory)


dholigan:
Thanks for upstreaming this. One heads-up on a gap we hit on the Ensemble E8 DK that carries over into the new MSPI driver.

The new `flash_mspi_is25xX0xx` / `mspi_dw_alif` path (like the HAL driver it replaces) enters memory-mapped XiP only at init and never re-enters it after a command-mode op, and there's no arbitration between XiP reads and indirect ops. On a part shared between XiP and a filesystem — or dual-core, where one core XiP-reads while the other issues indirect ops — the first indirect op after boot drops XiP, and every later memory-mapped read of the window bus-faults (precise data abort), with no diagnostic.

Minimal repro: build XiP-enabled, read the window (ok), do any `flash_write()`/`flash_erase()`, read the window again → fault.

Filed as ACS-2750 with the full write-up. We have a reference implementation (a runtime XiP-enable API plus a lock that holds XiP mode across a caller's critical section so indirect ops wait) and would be glad to contribute it against this driver — folded into this PR or as a follow-up, whichever you prefer. At minimum, an errata note that memory-mapped XiP is mutually exclusive with indirect access and must be re-armed after a command-mode op would help.

GaneshRamani-AlifSemi:
> Thanks for upstreaming this. One heads-up on a gap we hit on the Ensemble E8 DK that carries over into the new MSPI driver.
> 
> The new `flash_mspi_is25xX0xx` / `mspi_dw_alif` path (like the HAL driver it replaces) enters memory-mapped XiP only at init and never re-enters it after a command-mode op, and there's no arbitration between XiP reads and indirect ops. On a part shared between XiP and a filesystem — or dual-core, where one core XiP-reads while the other issues indirect ops — the first indirect op after boot drops XiP, and every later memory-mapped read of the window bus-faults (precise data abort), with no diagnostic.
> 
> Minimal repro: build XiP-enabled, read the window (ok), do any `flash_write()`/`flash_erase()`, read the window again → fault.
> 
> Filed as ACS-2750 with the full write-up. We have a reference implementation (a runtime XiP-enable API plus a lock that holds XiP mode across a caller's critical section so indirect ops wait) and would be glad to contribute it against this driver — folded into this PR or as a follow-up, whichever you prefer. At minimum, an errata note that memory-mapped XiP is mutually exclusive with indirect access and must be re-armed after a command-mode op would help.

Hi Darrell,

Thanks for the detailed feedback and reproduction steps.

This driver is currently under review and requires some additional restructuring to support safe coexistence between memory-mapped XiP access and command-based indirect operations.

Zephyr does not currently provide a standard API for arbitrating direct XiP access against indirect flash operations. For the new MSPI-based IS25 driver, we are working on the following approach:

- Serialize indirect flash operations and cooperating memory-mapped XiP readers using a common driver lock.
- Automatically reconfigure XiP after every indirect read, write or erase operation, before releasing that lock.
- Expose acquire/release APIs that applications can use around direct accesses to the memory-mapped XiP window.
- The proposed application-facing APIs are:

```
/**
 * @brief Acquire exclusive access to the memory-mapped XiP window.
 *
 * The caller must hold this lock for the complete duration of direct
 * memory-mapped accesses. Indirect flash operations use the same lock,
 * preventing them from changing the controller mode while a cooperating
 * XiP reader is active.
 *
 * This coordinates threads and SMP CPUs sharing one Zephyr kernel. It
 * does not coordinate cores running independent firmware images. Such
 * systems require a shared hardware semaphore ????.
 */
int flash_mspi_is25xx0xx_xip_acquire(const struct device *flash);

/**
 * @brief Release exclusive access to the memory-mapped XiP window.
 */
int flash_mspi_is25xx0xx_xip_release(const struct device *flash);
```
Applications that directly access the XiP address window must use this pair whenever indirect operations may occur concurrently. The driver will reconfigure XiP automatically after its indirect operations.

**This work is still under review and has not yet been implemented**. You mentioned that WHOOP has a reference implementation using runtime XiP configuration and a lock held across the critical section.

We would appreciate it if you could share that implementation or submit it as a PR for comparison.

Thanks !

dholigan:
[ospi-xip-arbitration-reference.patch](https://github.com/user-attachments/files/30242978/ospi-xip-arbitration-reference.patch)

dholigan:
@GaneshRamani-AlifSemi no problem, attached my patch

GaneshRamani-AlifSemi:
will address proper cherry-pick from upstream zephyr for mspi and is25wx drivers !

PR #594, opened 2026-07-19
drivers: audio: dmic: alif_pdm: recover from slab exhaustion instead of disabling IRQs

dholigan:
On a DMA slab-block allocation failure, the PDM warning-interrupt handler disabled all PDM interrupts, which permanently stopped capture — a consumer that briefly fell behind freeing blocks silently killed the microphone until a full peripheral re-init.

This reclaims the oldest queued block on exhaustion so capture continues (dropping oldest data instead of the whole stream), and adds real FIFO-overflow recovery with correct overflow-status handling.

Reproduced and fixed on the Ensemble E8 DK.

### DIFF
--- a/drivers/audio/alif_pdm.c
+++ b/drivers/audio/alif_pdm.c
@@ -33,6 +33,7 @@ struct pdm_data {
 	uint32_t slab_missed;
 	uint32_t record_data;
 	uint32_t bytes_got;
+	uint32_t overrun_count;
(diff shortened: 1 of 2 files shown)


Madhumathi-Alifsemi (commented):
Tested on the available boards. Looks good except for E7 LPPDM, which has an interrupt-routing issue with the warning/error interrupts.

PR #596, opened 2026-07-19
lib: libc: picolibc: select retarget-lock model by picolibc source

dholigan:
The picolibc retarget-lock model depends on which picolibc the build uses: the module picolibc uses the old `_LOCK_T = void *` model, while a toolchain picolibc (SDK 0.17.x, used when full libc is required) uses `struct __lock`. A wholesale move to `struct __lock` breaks the module build. This conditions the retarget-lock definitions on CONFIG_PICOLIBC_USE_MODULE so both picolibc sources build.

### DIFF
--- a/lib/libc/picolibc/locks.c
+++ b/lib/libc/picolibc/locks.c
@@ -7,6 +7,19 @@
 #include "picolibc-hooks.h"
 
 #ifdef CONFIG_MULTITHREADING
+
+/*
+ * Pick the picolibc retargetable-lock model that matches the active picolibc
+ * source. The pinned module picolibc uses the old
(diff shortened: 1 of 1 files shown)


sudhir-alifsemi (commented):
LGTM. 
The struct __lock implementation is effectively a backport of upstream Zephyr commit f075022beca4. Please add the upstream reference in the commit message description. Alternatively, it may be cleaner to cherry-pick the upstream commit first and then add a small follow-up patch to support older picolibc implementations using the feature macro check.

sudhir-alifsemi (approved):

RupeshKumar-AlifSemi (approved):

PR #598, opened 2026-07-22
drivers: i3c: dw: select role from dts

Shreehari-AlifSemi:
--Add a dts property to select the i3c controller's role during startup.

-- The current code just checks the Device Role register at startup and decides the mode. No option to change the role even if there is a provision from HW.

-- This commit allows the user to choose the mode as per requirement.

This PR is referred by https://github.com/alifsemi/sdk-alif/pull/874

### DIFF
--- a/drivers/i3c/i3c_dw.c
+++ b/drivers/i3c/i3c_dw.c
@@ -194,6 +194,7 @@ LOG_MODULE_REGISTER(i3c_dw, CONFIG_I3C_DW_LOG_LEVEL);
 #define SLV_CHAR_CTRL_IBI_REQUEST_CAPABLE  BIT(1)
 #define SLV_CHAR_CTRL_IBI_PAYLOAD          BIT(2)
 #define SLV_CHAR_CTRL_BCR_MASK             GENMASK(7, 0)
+#define SLV_CHAR_CTRL_DEVICE_ROLE_MASTER   (BIT(6) & GENMASK(7, 6))
 #define SLV_CHAR_CTRL_BCR(x)               ((x) & SLV_CHAR_CTRL_BCR_MASK)
 #define SLV_CHAR_CTRL_DCR_MASK             GENMASK(15, 8)
 #define SLV_CHAR_CTRL_DCR(x)               (((x) & SLV_CHAR_CTRL_DCR_MASK) >> 8)
@@ -240,7 +241,9 @@ LOG_MODULE_REGISTER(i3c_dw, CONFIG_I3C_DW_LOG_LEVEL);
 #define SLV_NACK_REQ_NACK_REQ_NACK 0x01
 
 #define DEVICE_CTRL_EXTENDED                           0xb0
-#define DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE(x)     ((x) & GENMASK(1, 0))
+#define DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_MASK   GENMASK(1, 0)
+#define DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE(x)     \
+	((x) & DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_MASK)
 #define DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_MASTER 0
 #define DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_SLAVE  1
 
@@ -366,6 +369,8 @@ struct dw_i3c_config {
 	const struct pinctrl_dev_config *pcfg;
 
 	void (*irq_config_func)();
+
+	bool target_mode;
 };
 
 struct dw_i3c_data {
@@ -2235,6 +2240,7 @@ static int dw_i3c_init(const struct device *dev)
 	uint32_t hw_capabilities;
 	uint32_t queue_capability;
 	uint32_t device_ctrl_ext;
+	uint32_t device_char_ctrl;
 
 	if (!device_is_ready(config->clock)) {
 		return -ENODEV;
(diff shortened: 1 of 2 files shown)


drivers/i3c/i3c_dw.c
```diff
@@ -2291,14 +2297,21 @@ static int dw_i3c_init(const struct device *dev)
...
 
-	/* if the boot condition starts as a target, then it's a secondary controller */
-	device_ctrl_ext = sys_read32(config->regs + DEVICE_CTRL_EXTENDED);
-	if (DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE(device_ctrl_ext) &
-	    DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_SLAVE) {
+	/* Set the controller's role Master/Secondary (Slave) */
+	device_ctrl_ext  = sys_read32(config->regs + DEVICE_CTRL_EXTENDED);
+	device_ctrl_ext &= ~DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_MASK;
+	device_char_ctrl = sys_read32(config->regs + SLV_CHAR_CTRL);
+	if (config->secondary_mode) {
```
sudhir-alifsemi: I dont think secondary_mode is the correct name here. Secondary can be a master as well. I would prefer to rename this property as target_mode for slave.

sudhir-alifsemi (changes requested):

drivers/i3c/i3c_dw.c
Shreehari-AlifSemi: updated

drivers/i3c/i3c_dw.c
```diff
@@ -2291,14 +2297,21 @@ static int dw_i3c_init(const struct device *dev)
...
-	if (DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE(device_ctrl_ext) &
-	    DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_SLAVE) {
+	/* Set the controller's role Master/Secondary (Slave) */
+	device_ctrl_ext  = sys_read32(config->regs + DEVICE_CTRL_EXTENDED);
+	device_ctrl_ext &= ~DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_MASK;
+	device_char_ctrl = sys_read32(config->regs + SLV_CHAR_CTRL);
+	if (config->target_mode) {
+		device_ctrl_ext  |= DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_SLAVE;
+		device_char_ctrl &= ~SLV_CHAR_CTRL_DEVICE_ROLE_MASTER;
 		ctrl_config->is_secondary = true;
```
sudhir-alifsemi, line 2307: Shouldnt we set is_secondary to false for the target_mode?

Shreehari-AlifSemi: No, the reason being the current slave is capable of becoming the master (Primary controller as our dw i3c is capable of it) of the bus if bus master accepts. So, it is a secondary controller when in slave mode.

drivers/i3c/i3c_dw.c
```diff
@@ -2291,14 +2297,21 @@ static int dw_i3c_init(const struct device *dev)
...
 	} else {
+		device_ctrl_ext |= DEVICE_CTRL_EXTENDED_DEV_OPERATION_MODE_MASTER;
+		device_char_ctrl |= SLV_CHAR_CTRL_DEVICE_ROLE_MASTER;
 		ctrl_config->is_secondary = false;
 	}
```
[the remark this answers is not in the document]
Shreehari-AlifSemi: SLV_CHAR_CTRL_DEVICE_ROLE_MASTER is doing the same

drivers/i3c/i3c_dw.c
```diff
@@ -2406,6 +2419,7 @@ static DEVICE_API(i3c, dw_i3c_api) = {
 		.common.dev_list.num_i3c = ARRAY_SIZE(dw_i3c_device_array_##n),                    \
 		.common.dev_list.i2c = dw_i3c_i2c_device_array_##n,                                \
 		.common.dev_list.num_i2c = ARRAY_SIZE(dw_i3c_i2c_device_array_##n),                \
+		.target_mode = DT_INST_PROP(n, target_mode),                                 \
```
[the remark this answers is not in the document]
Shreehari-AlifSemi: U need DT_INST_PROP_OR for properties other than boolean. It takes 1 if boolen property is defined 0 otherwise.

drivers/i3c/i3c_dw.c
sudhir-alifsemi: As I understand from databook, I3C can operate as "Dual Role Device -
is a digital controller that implements all protocol functions related to Controller (Active Controller and Secondary Controller) and Target as defined in the MIPI I3C Specification". The word secondary is actually the controller which can act as a master and not as slave.

Shreehari-AlifSemi: Yes, dual role but there should be only one bus master at a time on the bus. So, the secondary controller is automatically a slave at that point unless he got the bus mastership.

sudhir-alifsemi: Understood. Based on the databook and the I3C specification, a Secondary Controller may initially behave as a Target on the bus, but it remains controller-capable and can become the Active Controller through controller handoff.

sudhir-alifsemi (approved):

drivers/i3c/i3c_dw.c
```diff
@@ -194,6 +194,7 @@ LOG_MODULE_REGISTER(i3c_dw, CONFIG_I3C_DW_LOG_LEVEL);
 #define SLV_CHAR_CTRL_IBI_REQUEST_CAPABLE  BIT(1)
 #define SLV_CHAR_CTRL_IBI_PAYLOAD          BIT(2)
 #define SLV_CHAR_CTRL_BCR_MASK             GENMASK(7, 0)
+#define SLV_CHAR_CTRL_DEVICE_ROLE_MASTER   (BIT(6) & GENMASK(7, 6))
```
silesh-alifsemi, line 197: `GENMASK(7, 6)` is a nop here and this is always just `BIT(6) `? If this is a single bit field in the register, `BIT(6) ` is enough.

silesh-alifsemi (changes requested):

drivers/i3c/i3c_dw.c
Shreehari-AlifSemi: Updated

silesh-alifsemi (approved):

PR #605, opened 2026-07-31
Add Usb hid sample support for Alif

Deepak-AlifSemi:
Add USB HID joystick sample using input subsystem and gpio-keys,
with Alif board-specific snippet overlays and configurations.
Place UDC static buffer for HID joystick in .alif_ns section for
both ensemble and balletto SoC linker scripts.


RupeshKumar-AlifSemi (approved):
Looks OK

silesh-alifsemi (approved):

